PR #1, opened 2016-11-04
added files generated by provider generator

durandom:
Files in here are 100% generated by the provider generator.
They should make the extraction PR https://github.com/ManageIQ/manageiq/pull/12274 green.

If you find anything odd or not good looking I try to fix it in the generator

@Fryguy @djberg96 @bronaghs @juliancheal @blomquisg

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -1,0 +1,8 @@
+begin
+  require 'bundler/setup'
+rescue LoadError
+  puts 'You must `gem install bundler` and `bundle install` to run rake tasks'
+end
+
+APP_RAKEFILE = File.expand_path("../spec/manageiq/Rakefile", __FILE__)
+load 'rails/tasks/engine.rake'
--- a/lib/manageiq/providers/azure.rb
+++ b/lib/manageiq/providers/azure.rb
@@ -1,0 +1,8 @@
+require "manageiq/providers/azure/engine"
+
+module Manageiq
+  module Providers
+    module Azure
+    end
+  end
+end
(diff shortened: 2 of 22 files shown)


chessbyte:
@Fryguy @durandom was Travis supposed to run here?

durandom:
I guess travis is not setup yet.

juliancheal:
@durandom You may as well add the commits from https://github.com/ManageIQ/manageiq-providers-amazon/commit/536ce7a2ba6dfc04cc832c560731bbb335825d36

As Codeclimate has changed.

Rakefile
```diff
@@ -0,0 +1,8 @@
+begin
+  require 'bundler/setup'
+rescue LoadError
+  puts 'You must `gem install bundler` and `bundle install` to run rake tasks'
```
djberg96: Do you want to re-raise the error here? Or exit? Also, maybe substitute `puts` with `warn`.

djberg96: Then it needs work ;)

djberg96:
I vote to merge. Let's get going.

lib/manageiq/providers/azure.rb
```diff
@@ -0,0 +1,8 @@
+require "manageiq/providers/azure/engine"
+
+module Manageiq
```
Fryguy: Minor typo, but this should be `ManageIQ`

Fryguy: Going to merge as is... @durandom Please fix this in a follow up PR and in the generator as well

durandom: Oops, are modules case insensitve?
Will fix this...

PR #2, opened 2016-11-08
Define runner for all workers

kbrock:
There is currently a requirement that all workers have a runner.
Also added missing loading for our event_catchers

Making this consistent across projects

This was split off of https://github.com/ManageIQ/manageiq/pull/12461


kbrock:
having trouble running tests.
Any suggestions?

djberg96:
@kbrock Looks like it's green now.

Looks like we ought to merge.

PR #3, opened 2016-11-08
chmod 755 bin/setup

kbrock:
setup needs to be executable


durandom:
will hold off merging, until we get travis here.

I'll change this in the provider generator too, good catch!

Also, while at it, could you also chmod bin/rails ?

kbrock:
ok, done

I never got the tests running locally. so I'm guessing there is some other work to do in `bin/setup`

durandom:
@kbrock tests work for me with `bin/rails app:test:providers:azure` - The tricky part is, they run under the `app:` namespace, because engines need a test app

Fryguy:
Will have it shortly... @jrafanie is working on the CodeClimate issues.

kbrock:
kicking

durandom:
@kbrock thanks for the PR but this came in in https://github.com/ManageIQ/manageiq-providers-azure/pull/8 already. I could not get your PR to get :green_apple: because of the e6shim thing

PR #4, opened 2016-11-08
Set Azure log to $azure_log

djberg96:
Set the Azure log to $azure_log. This will log the http requests being made by the azure-armrest gem.

Possible issues include the fact that there's a single, global log for rest-client which could be overwritten, and the fact that I cannot currently apply a filter to the Logger instance because of https://github.com/rest-client/rest-client/pull/543.


djberg96:
@miq-bot assign @blomquisg

PR #5, opened 2016-11-09
remove not needed module with incorrect spelling

durandom:
this sneaked in somehow.

azure/engine.rb is defining the modules already.


djberg96:
👍

durandom:
already fixed in https://github.com/ManageIQ/manageiq-providers-azure/pull/8

PR #8, opened 2016-11-10
align files with output from rails generate provider Azure

durandom:
ran `rails generate provider Azure` again


durandom:
@Fryguy let's merge this. Hopefully this will turn the other PRs green - because this also `chmod 755 bin/rails`

PR #9, opened 2016-11-10
Initial setup of locale files for ManageIQ::Providers::Azure

mzazrivec:
These files were created by a rake task:

    rake locale:plugin:find[ManageIQ::Providers::Azure]

run from ManageIQ git checkout.

### DIFF
--- a/locale/ManageIQ_Providers_Azure.pot
+++ b/locale/ManageIQ_Providers_Azure.pot
@@ -1,0 +1,155 @@
+# Gettext catalog for ManageIQ Azure provider plugin
+# Copyright (C) 2016
+# This file is distributed under the same license as the ManageIQ_Providers_Azure package.
+# Milan Zázrivec <mzazrivec@redhat.com>, 2016.
+#
+#, fuzzy
+msgid ""
+msgstr ""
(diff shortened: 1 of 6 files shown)


locale/ManageIQ_Providers_Azure.pot
```diff
@@ -0,0 +1,155 @@
+# SOME DESCRIPTIVE TITLE.
+# Copyright (C) YEAR THE PACKAGE'S COPYRIGHT HOLDER
+# This file is distributed under the same license as the ManageIQ_Providers_Azure package.
+# FIRST AUTHOR <EMAIL@ADDRESS>, YEAR.
```
Fryguy: Should these be updated?

mzazrivec: Hmm, don't know. Things will work just fine even without this info being filled in :smile: 

I'd say this is just a cosmetic issue, but I can definitely fix this if it's important.

djberg96: I worry that if you don't update it now it will never get updated. :)

mzazrivec: @djberg96 Alrighty, you convinced me :smile:

djberg96:
👍

PR #16, opened 2016-11-30
Move from MIQ repo to provider repo.

djberg96:
Just moving this over to our provider-specific repo.

@durandom Is there anything else that needs to happen for this to be picked up?

### DIFF
--- a/content/miq_dialogs/miq_provision_azure_dialogs_template.yaml
+++ b/content/miq_dialogs/miq_provision_azure_dialogs_template.yaml
@@ -1,0 +1,383 @@
+--- 
+:name: miq_provision_azure_dialogs_template
+:description: Sample Azure Instance Provisioning Dialog
+:dialog_type: MiqProvisionWorkflow
+:content: 
+  :buttons: 
+  - :submit
+  - :cancel
+  :dialogs: 
+    :requester: 
+      :description: Request
+      :fields: 
+        :owner_phone: 
+          :description: Phone
+          :required: false
+          :display: :hide
+          :data_type: :string
+        :owner_country: 
+          :description: Country/Region
+          :required: false
+          :display: :hide
+          :data_type: :string
+        :owner_phone_mobile: 
+          :description: Mobile
+          :required: false
+          :display: :hide
+          :data_type: :string
+        :owner_title: 
+          :description: Title
+          :required: false
+          :display: :hide
+          :data_type: :string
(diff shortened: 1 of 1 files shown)


durandom:
@djberg96 oh, I have _zero_ knowledge about those product files :( 
I'm not sure what picks those files up.
Is that automates domain?

gmcculloug:
These files get seeded in the database from [app/models/miq_dialog.rb] (https://github.com/ManageIQ/manageiq/blob/28ea1414befb2aee4152dba397d04250b93b6bfd/app/models/miq_dialog.rb#L16).

Have we updated any seeding methods to deal with pluggable providers?  Is there a pattern for this yet?

durandom:
@gmcculloug 
> Have we updated any seeding methods to deal with pluggable providers? Is there a pattern for this yet?

Seeding as in [EvmDatabase.seed](https://github.com/durandom/manageiq/blob/17871535a1ff95f69efab3121da89ac56e544078/lib/evm_database.rb#L55) just works ™️  because of descendent_loader.

For loading files, we havent moved anything yet 😢 
But we have [Vmdb::Plugins](https://github.com/ManageIQ/manageiq/blob/master/lib/vmdb/plugins.rb) that is used in @bdunne https://github.com/ManageIQ/manageiq/pull/11083

@bdunne am I right, that once https://github.com/ManageIQ/manageiq/pull/11083 is merged this could act as a model for seeding dialogs as well?

bdunne:
@durandom That's a different seed for the dialogs.  Check out [MiqDialog.seed](https://github.com/ManageIQ/manageiq/blob/master/app/models/miq_dialog.rb#L21).

durandom:
> could act as a model for seeding

Oh, I meant to act as a role model, like a pattern for doing this kind of file seeding.
So once https://github.com/ManageIQ/manageiq/pull/11083 is merged we should look into how the dialog files are seeded

durandom:
Added [WIP] as we dont have a way of seeding dialogs yet

djberg96:
@durandom ManageIQ/manageiq#11083 was merged. Should we revisit?

bdunne:
@djberg96 I think the dialog files should live under `content/dialogs`.  Still, we need the Dialog seeder to know about plugins first.

djberg96:
@durandom Done.

https://github.com/ManageIQ/manageiq/pull/14655

PR #17, opened 2016-12-07
Load balancer networking

bronaghs:
In order to determine which Floating Ips are available for recycling during instance provisioning, we need to consider that some may be associated with Load balancers, not just instances.

This PR associates a Floating Ip with its Load balancer through a fake network port; a necessary step given that one is not already present.

https://bugzilla.redhat.com/show_bug.cgi?id=1380728

### DIFF
--- a/app/models/manageiq/providers/azure/network_manager/floating_ip.rb
+++ b/app/models/manageiq/providers/azure/network_manager/floating_ip.rb
@@ -1,5 +1,2 @@
 class ManageIQ::Providers::Azure::NetworkManager::FloatingIp < ::FloatingIp
-  def self.available
-    joins(:network_port).where({ "network_ports.device_id" => nil })
-  end
 end
--- a/app/models/manageiq/providers/azure/network_manager/floating_ip.rb
+++ b/app/models/manageiq/providers/azure/network_manager/floating_ip.rb
@@ -1,2 +1,5 @@
 class ManageIQ::Providers::Azure::NetworkManager::FloatingIp < ::FloatingIp
+  def self.available
+    joins(:network_port).where({ "network_ports.device_id" => nil })
+  end
 end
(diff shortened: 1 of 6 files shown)


bronaghs:
@miq-bot add_label wip

app/models/manageiq/providers/azure/network_manager/floating_ip.rb
```diff
@@ -1,2 +1,7 @@
 class ManageIQ::Providers::Azure::NetworkManager::FloatingIp < ::FloatingIp
+  belongs_to :load_balancer
+
+  def self.available
+    where(:vm_id => nil, :load_balancer_id => nil)
```
Ladas: does Azure document somewhere what entities can have a floating IP?

bronaghs: Hey @ladas - https://msdn.microsoft.com/en-us/library/azure/mt163638.aspx - see paragraph 2

Ladas: right, seems like only LB and VM can be associated here, so this should be good for now

Ladas:
@bronaghs this should be good for now :+1: 

I am thinking, for cross providers, it could be enough to store the connected device as a text (e.g. device id column) and base 'available?' on that being nil?

Then this should work for all providers and also for objects we don't track yet, since we will see there is some id, but we can't link it in our DB.

Ladas:
@miq-bot assign @blomquisg

Ladas:
@blomquisg we can merge this, to tackle the current issue. Then as a follow up, we can see if there is a cross-provider solution.

bronaghs:
@Ladas - can you take a look while I write the specs?

djberg96:
@Ladas @bronaghs Sorry about the conflicting refresher.yml, @durandom had to regenerate the cassettes for one of his PR's. In theory you should be able to just use the new one.

bronaghs:
@Ladas - please review

Ladas:
@bronaghs looks great

PR #18, opened 2016-12-09
split rubocop files for dependency on manageiq

durandom:
* changes rubocop config to point to manageiq repo instead of manageiq/guides
* don't maintain our own copy of rubocop_cc.yml for codeclimate, but use their support for [external configurations](https://codeclimate.com/changelog/582495c32c33066f1b00191d) 

following along https://github.com/ManageIQ/manageiq-providers-amazon/pull/90


djberg96:
@durandom Quick question, will spec/manageiq/.rubocop.yml need any adjustment?

durandom:
@djberg96 `spec/manageiq/.rubocop.yml` is not used at all. All under `spec/manageiq` is just the dummy app to run specs and pulls in dependencies from the `spec/manageiq/Gemfile`

djberg96:
@durandom Ok, thanks for the clarification.

durandom:
@bronaghs can you have a look and merge?

.rubocop_local.yml
```diff
@@ -0,0 +1,4 @@
+GlobalVars:
+  AllowedVariables:
+  # Loggers
+  - $log
```
bronaghs: @durandom - looks like the azure_log be added here too, right?

durandom:
@bronaghs $azure_log is in. Good catch.

PR #20, opened 2016-12-20
Ignore config/secrets.yml

djberg96:
This mirrors the original Rails repo so we can store our own credentials in config/secrets.yml. This is used mostly for when we need to regenerate VCR cassettes.


bronaghs:
👍 LGTM @djberg96

djberg96:
@bronaghs Eh, I guess that's not the place for it anyway. It should go under spec/manageiq/config. This change was harmless, though.

PR #21, opened 2016-12-21
Azure deployment refresh enhancement

bzwei:
1. Use the recently added Azure API to directly get deployment template. With old API we can only retrieve deployment template if the URI is known, or if the deployment is provisioned from MangeIQ. As a result many deployments cannot recover their templates. With the newest API we now can directly get the templates without downloading from an URI or loading from ManageIQ DB. But the API does not work with old deployments, so we have to keep existing logic as the fallback method.

2. Simplify how to identify resources created by a deployment. We now understand the key to identify them is provisioning_operation == 'Create'. This allows us to delete quite some code and simply the logic.

It seems above improvements should be separate PRs; but they all require regenerating the VCR for refresh spec test. In addition, our recent upgrade `azure-armrest` gem version to `0.5.0` also requires  regenerating the VCR. The work to regenerate VCR is non-trivial, so that I would like to put them together with one commit.

This PR has a dependency on https://github.com/ManageIQ/manageiq-gems-pending/pull/30


bzwei:
/cc @blomquisg @djberg96 @bronaghs
The build failed because we need a newer azure-armrest. ManageIQ/manageiq-gems-pending#30 needs to be merged first.

djberg96:
@bzwei Sorry for the conflicting files, @durandom had to update the cassettes for one of his PR's.

djberg96:
@bzwei Should work now with https://github.com/ManageIQ/manageiq-providers-azure/pull/28 merged.

app/models/manageiq/providers/azure/cloud_manager/refresh_parser.rb
```diff
@@ -1,3 +1,5 @@
+require 'azure-armrest'
+
```
djberg96: I've learned the hard way that we shouldn't do this. It causes a load ordering problem that results in the Rails cache getting cleared because of the cache_method gem:

https://github.com/ManageIQ/manageiq-providers-azure/pull/38

djberg96:
You should be able to just use the cassettes and specs that are already there.

bzwei:
@miq-bot assign @blomquisg 
@miq-bot add_label enhancement
@miq-bot add_label euwe/no

bzwei:
With the new API support we are now able to get templates for many stacks that we used cannot due to the lacking of uri. For this reason we see the number of total orchestration templates in the testing provider increases from 4 to 18.

blomquisg:
@djberg96 can you give me a :+1: on this one?

djberg96:
👍

PR #22, opened 2017-01-09
Set default rake task

djberg96:
This just sets a default task so I can do `bundle exec rake` for this repo and run the tests.


djberg96:
Looks like it was set by https://github.com/ManageIQ/manageiq-providers-azure/pull/27.

Closing.

PR #24, opened 2017-01-16
Delete all resources when deleting an Azure stack

bzwei:
Azure API by default does not delete associated resources of a stack. In `azure-armrest` we made it possible to delete all associated resources together with a deployment.

We now make this default behavior for upstream, but for previous versions we should make it only optional in order not to surprise existing users. Therefore this change should not be auto back ported.

@djberg96 please review.


bzwei:
@miq-bot add_label euwe/no

bzwei:
@miq-bot remove_label euwe/no
Will need to back port to euwe and darga, but through separate PRs.

djberg96:
👍

simaishi:
Backported to Darga via https://github.com/ManageIQ/manageiq/pull/13547

simaishi:
Backported to Euwe via https://github.com/ManageIQ/manageiq/pull/13548

PR #25, opened 2017-01-16
Return the available Floating IPs in Azure.

bronaghs:
PR [#12300](ManageIQ/manageiq#12300) changed the Floating::IP available method to return floating IPs not associated with either a VM or a NIC. The FloatingIp::available method is used in cloud instance provisioning to display a list of available floating IPs that can be re-used. This change inadvertently affected the Azure provider integration where a FloatingIp is considered available when it is associated with a NIC as long as that NIC is not attached to another device, which can occur when a VM is deleted and its NIC and IP remain orphaned.

This PR is for an Azure specific way to determine available Floating Ips.

https://bugzilla.redhat.com/show_bug.cgi?id=1380728

### DIFF
--- a/app/models/manageiq/providers/azure/network_manager/floating_ip.rb
+++ b/app/models/manageiq/providers/azure/network_manager/floating_ip.rb
@@ -1,2 +1,5 @@
 class ManageIQ::Providers::Azure::NetworkManager::FloatingIp < ::FloatingIp
+  def self.available
+    joins(:network_port).where("network_ports.device_id" => nil) + where(:network_port_id => nil)
+  end
 end
--- a/app/models/manageiq/providers/azure/network_manager.rb
+++ b/app/models/manageiq/providers/azure/network_manager.rb
@@ -17,6 +17,9 @@ class ManageIQ::Providers::Azure::NetworkManager < ManageIQ::Providers::NetworkM
 
   include ManageIQ::Providers::Azure::ManagerMixin
 
+  has_many :floating_ips, :foreign_key => :ems_id, :dependent => :destroy,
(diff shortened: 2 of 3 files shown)


bronaghs:
@djberg96, @blomquisg  - please review.

app/models/manageiq/providers/azure/network_manager/floating_ip.rb
```diff
@@ -1,2 +1,5 @@
 class ManageIQ::Providers::Azure::NetworkManager::FloatingIp < ::FloatingIp
+  def self.available
+    joins(:network_port).where({ "network_ports.device_id" => nil }) + where(:network_port_id => nil)
+  end
```
bronaghs: @djberg96 @blomquisg - expensive? I'm not sure how to optimize this given the associations that are in place.

app/models/manageiq/providers/azure/network_manager/floating_ip.rb
```diff
@@ -1,2 +1,5 @@
 class ManageIQ::Providers::Azure::NetworkManager::FloatingIp < ::FloatingIp
+  def self.available
```
djberg96: Should this be `:available?`?

bronaghs: @djberg96 - I had wondered the same thing, but no:
https://github.com/ManageIQ/manageiq/blob/master/app/models/floating_ip.rb

djberg96: @bronaghs ok, understood.

djberg96:
👍

PR #32, opened 2017-02-07
Setup database.yml and v2_key in bin/setup

djberg96:
This is just to make initial setup a little nicer. It's easy to forget to set these up before attempting to run tests.

### DIFF
--- a/bin/setup
+++ b/bin/setup
@@ -1,11 +1,14 @@
 #!/usr/bin/env ruby
 require 'pathname'
+require 'fileutils'
 
 gem_root = Pathname.new(__dir__).join("..")
 
 unless gem_root.join("spec/manageiq").exist?
   puts "== Cloning manageiq sample app =="
   system "git clone https://github.com/ManageIQ/manageiq.git --branch master --depth 1 spec/manageiq"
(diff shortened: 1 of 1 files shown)


bin/setup
```diff
@@ -1,11 +1,14 @@
...
 require 'pathname'
+require 'fileutils'
 
 gem_root = Pathname.new(__dir__).join("..")
 
 unless gem_root.join("spec/manageiq").exist?
   puts "== Cloning manageiq sample app =="
   system "git clone https://github.com/ManageIQ/manageiq.git --branch master --depth 1 spec/manageiq"
+  FileUtils.cp "spec/manageiq/config/database.pg.yml", "spec/manageiq/config/database.yml", :verbose => true
+  FileUtils.cp "spec/manageiq/certs/v2_key.dev", "spec/manageiq/certs/v2_key", :verbose => true
```
bdunne: This isn't necessary since it's already covered at the end of this file by `ManageIQ::Environment.manageiq_plugin_setup`

See https://github.com/ManageIQ/manageiq/blob/master/lib/manageiq/environment.rb#L12

djberg96: @bdunne Oh, you're right. I must've cloned it manually before and didn't realize the bin/setup did it automatically.

PR #33, opened 2017-02-07
Add development instructions.

djberg96:
This just adds some "quick dev" instructions to the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,9 @@ ManageIQ plugin for the Azure provider.
 
 ## Development
 
-See the section on pluggable providers in the [ManageIQ Developer Setup](http://manageiq.org/docs/guides/developer_setup)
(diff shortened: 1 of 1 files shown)


durandom:
TBH I'm not a big fan of duplicated docs as they tend to go out of sync, but this is probably good for a quick heads up. 

@djberg96 bonus points up for grabs for adding that to the generator as well :)

bdunne:
@durandom Are you suggesting linking to a centralized doc on ManageIQ/guides?  I would :+1: that.

durandom:
> Are you suggesting linking to a centralized doc on ManageIQ/guides? I would 👍 that.

well, the link is already in the README and links to http://manageiq.org/docs/guides/developer_setup

Cant link to markdown lines :(

PR #34, opened 2017-02-07
Add update task, and automatically update before running specs.

djberg96:
This PR adds an `:update` task, which is just a wrapper for "bin/update".

Also, since it's easy to forget to update the core repo before running specs, which can lead to unexpected errors, the :update task was added as a prerequisite to the `:spec` task.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -14,7 +14,12 @@ namespace :spec do
   task :setup => 'app:test:providers:azure:setup'
 end
 
+desc "Update your local ManageIQ repository"
+task :update do
+  sh "bin/update" unless ENV['CI']
+end
+
 desc "Run all azure specs"
-task :spec => 'app:test:providers:azure'
+task :spec => [:update, 'app:test:providers:azure']
 
 task :default => :spec


durandom:
I dont think this is necessary. I effect this means you run git pull every time you run the specs.
But then, I dont use this very often anyway - so it might be a time saver for others.

But, as with your README update, if this makes it into here, I'd like to see it in the provider generator too, so we keep the providers in sync as much as possible

djberg96:
@durandom If you merge it, I will update this PR:

https://github.com/ManageIQ/manageiq/pull/13817

durandom:
@bdunne could you give your opinion here?

bdunne:
@durandom I think I'm in the same position as you, I don't think it's necessary.  I don't often work on the plugin repos, but when I do I usually have `spec/manageiq` symlinked to my clone of `manageiq`, so the `bin/update` would be a no-op and just consume more time.

durandom:
ok, how about merging it here and see how it pans out?

PR #35, opened 2017-02-08
Give orchestration stack double a name

djberg96:
This is a cherry-pick from the core repo.

See https://github.com/ManageIQ/manageiq/pull/13770


bdunne:
:+1: LGTM Would merge if I could

PR #36, opened 2017-02-09
Disable metrics and events in unsupported regions

djberg96:
This PR uses the SupportsFeatureMixin and the most recent azure-armrest gem to disable metrics and events collection in regions that are not supported.

Note that this requires the addition of `:metrics` to the support matrix. See https://github.com/ManageIQ/manageiq/pull/13381

~~You will also have to update gems-pending-config to point to azure-armrest 0.6.0.~~

### DIFF
--- a/app/models/manageiq/providers/azure/cloud_manager.rb
+++ b/app/models/manageiq/providers/azure/cloud_manager.rb
@@ -1,3 +1,5 @@
+require 'azure-armrest'
+
 class ManageIQ::Providers::Azure::CloudManager < ManageIQ::Providers::CloudManager
   require_nested :AvailabilityZone
   require_nested :EventCatcher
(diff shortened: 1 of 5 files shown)


djberg96:
@miq-bot add_label wip

djberg96:
Note to self: as per conversation with the UI folks, we may want to add a textual summary for the UI at app/helpers/ems_cloud_helper/textual_summary.rb (in the classic-ui repo).

djberg96:
@miq-bot remove_label wip

djberg96:
@bronaghs @durandom Please take a look. The event handling is very similar to the OpenStack approach. For metrics, I just put a check within the `get_counters` method which basically short circuits the issue until a more unified solution comes along.

app/models/manageiq/providers/azure/cloud_manager/metrics_capture.rb
```diff
@@ -183,6 +183,13 @@ def raw_metrics_for_counter(storage_conn, counter, start_time, end_time)
   end
 
   def get_counters(metrics_conn)
+    ems = target.ext_management_system
+
+    unless ems.supports_timeline?
```
durandom: would this be better `ems.insights?` to not conflate with timelines / events ?

app/models/manageiq/providers/azure/cloud_manager.rb
```diff
@@ -1,3 +1,5 @@
+require 'azure-armrest'
```
djberg96: I was hitting some kind of weird load ordering issue that I couldn't nail down, which resulted in undefined constant errors.

app/models/manageiq/providers/azure/cloud_manager/metrics_capture.rb
djberg96: Yes, probably. I will update.

bronaghs: @djberg96 - I would probably move this to line 78 after we check for an EMS, this would keep the checks together and avoid calling methods if metrics isn't supported. Or is there a reason for placing this line here?

djberg96: I'll try it and report back.

djberg96: @bronaghs Ok, I've moved it as you suggested. Seems to work just as well there.

djberg96: @bronaghs Upon further review it seems to cause test failures because of with_provider_connection which is called again later within the same method. I've moved it back for now.

PR #51, opened 2017-04-13
Refactor service creation, lock down api-version strings

djberg96:
The primary goal of this PR is to lock down `api-version` string dates for use with the Azure REST API. This string is passed along with every HTTP request made to Azure. Currently we default to the latest api-version string for any given service provider. Normally this is fine, but on rare occasions Microsoft does change the resulting output. This has already happened at least twice already, once for storage accounts and once for metrics definitions.

The problem is that if the Azure output changes, our autogenerated model and its corresponding methods could potentially break. Consequently, I feel it that it's best if we lock down the api-version string for any service class that we use via the `config/settings.yml` file. Each service then has its api-version set explicitly when constructed, and we know it won't break for any particular major release of ManageIQ.

I've also taken the liberty of refactoring the code to use a shared set of methods in the RefresherHelper module, since we had code duplication between the Network Manager and the Cloud Manager. This should provide a small speed boost, too.


bronaghs:
LGTM

app/models/manageiq/providers/azure/refresh_helper_methods.rb
```diff
@@ -57,4 +57,70 @@ def network_interfaces
...
+
+  # Create the necessary service classes and lock down their api-version
+  # strings using the config/settings.yml from the provider repo. The
+  # "to_s" call for the version strings puts the date in the format
+  # that we need, i.e. "YYYY-MM-DD".
+  #
+
+  def availability_set_service(config)
+    @avs ||= ::Azure::Armrest::AvailabilitySetService.new(config)
+    @avs.api_version = Settings.ems.ems_azure.api_versions.availability_set.to_s
```
durandom: why `.to_s` ? will `Settings` parse it into a `Date`?

app/models/manageiq/providers/azure/refresh_helper_methods.rb
```diff
@@ -57,4 +57,70 @@ def network_interfaces
...
   end
+
+  # Create the necessary service classes and lock down their api-version
+  # strings using the config/settings.yml from the provider repo. The
+  # "to_s" call for the version strings puts the date in the format
+  # that we need, i.e. "YYYY-MM-DD".
+  #
+
+  def availability_set_service(config)
+    @avs ||= ::Azure::Armrest::AvailabilitySetService.new(config)
```
durandom: isnt `||=` dangerous here? Because you'll might get a memoized version although you provide a different `config`

From a quick glance I would just drop memoiziation and instance vars. Or maybe you can set `api_version` in config?!

```ruby
def availability_set_service(config)
  ::Azure::Armrest::AvailabilitySetService.new(config).tap do |a|
    a.api_version = Settings.ems.ems_azure.api_versions.availability_set.to_s
  end
end
```

durandom:
Besides those questions, looks great! And cool to see more provider specific settings :)

app/models/manageiq/providers/azure/refresh_helper_methods.rb
djberg96: The config has a default api-version, but individual services have their own api-version strings. The config value is only used as a fallback, since we don't really want its value applied globally.

I always forget about tap/do. Thanks, I'll update that.

As for memoized vars, I suppose that could be an issue for multiple Azure providers, but I need to double check.

app/models/manageiq/providers/azure/refresh_helper_methods.rb
djberg96: Yes, the YAML lib converts date strings to Date objects. Calling `.to_s` puts it in the format we need.

app/models/manageiq/providers/azure/refresh_helper_methods.rb
Ladas: the memoized values look safe

Ladas:
looks great, I am fine with not having the tap :-)

app/models/manageiq/providers/azure/refresh_helper_methods.rb
durandom: > the memoized values look safe

It's safe because you know how it's used in the code. Until somebody comes along...

I would assume this method returns a service based on the config but not set an instance variable.

durandom:
I dont care about `tap` either 🤷‍♂️  It's just some fancy style. But I think the memoization in this method is wrong. It's also wrong to set instance variables in helper modules, but this would start a whole other discussion 😛

app/models/manageiq/providers/azure/refresh_helper_methods.rb
Ladas: right, but this is like 1 config per 1 refresh right? Or do I miss something? :-)

djberg96:
@durandom I actually plan on refactoring them out in a future PR. I actually started doing that, but then decided it was getting too huge.

durandom:
K, all I'm saying the code could be a bit _cleaner_ if it was a method without a side effect. 
But you can merge if you like it that way

djberg96:
@durandom @Ladas Ok, I've removed the memoization and instance variable creation. @bzwei I've changed it to use `.tap` as well.

djberg96:
@durandom Hm, we should be able to backport this to both Fine and Euwe shouldn't we?

durandom:
yeah, but you probably need to make dedicated PRs, because the settings.yml wont cherry-pick

djberg96:
@durandom Should work for Fine, though, right?

bronaghs:
@djberg96 - I think think this should be find/yes, right? See comment [here](https://github.com/ManageIQ/manageiq-providers-azure/pull/65#issuecomment-300269580)

djberg96:
@bronaghs Yes.

djberg96:
@miq-bot remove_label fine/no

djberg96:
@miq-bot add_label fine/yes

simaishi:
Backported to Euwe via https://github.com/ManageIQ/manageiq/pull/15056

simaishi:
BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1439815

simaishi:
Fine backport details:

```
$ git log -1
commit 2864c7bf35dab09955cdf51f87455b841d5dd2f8
Author: Marcel Hild <hild@b4mad.net>
Date:   Mon Apr 24 20:08:59 2017 +0200

    Merge pull request #51 from djberg96/api_versions2
    
    Refactor service creation, lock down api-version strings
    (cherry picked from commit 2aefc8dc7869ce26fcb7c38ea255f8760663cf41)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1459217
```

PR #55, opened 2017-04-20
Make azure-armrest dependency less pessimistic

djberg96:
This lines up with how the dependencies for gems-pending is setup, and is generally easier to deal with.

### DIFF
--- a/manageiq-providers-azure.gemspec
+++ b/manageiq-providers-azure.gemspec
@@ -13,7 +13,7 @@ Gem::Specification.new do |s|
 
   s.files = Dir["{app,config.lib}/**/*"]
 
-  s.add_dependency "azure-armrest", "=0.7.1"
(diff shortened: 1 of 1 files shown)


bronaghs:
This means we could point to a new release of the azure-armrest gem before we have updated the manageiq-providers-azure code to be compatible with it. We got burned by this a while ago and have since been explicit. Last I heard we had to be explicit with all gems developed by ManageIQ.

djberg96:
@bronaghs True, but we were a bit careless in the past with compatibility and versioning. We've been pretty good about avoiding that for a while now.

bronaghs:
@blomquisg - what's your take?

bdunne:
If `azure-armrest` follows [Semver](http://semver.org/), everything should be Fine. :wink:

Issue #46, opened 2017-04-12
move service dialogs from core


durandom:
https://github.com/ManageIQ/manageiq/pull/14668 has been merged. Now we can also seed `content/service_dialogs`. Looks like there is at least one for azure.

@djberg96 do you want to move this over here?

djberg96:
@durandom I'll get a PR in today. What about `orchestration_templates`?

durandom:
@djberg96 cool. 

> I'll get a PR in today. What about orchestration_templates?

Dont know :) Are they seeded like the dialogs?

djberg96:
@bdunne Do you know the answer to that?

bdunne:
Looks like seeding for that model needs to be updated for plugins as well.  I would expect them to move to `content/orchestration_templates/`.  cc @bzwei

durandom:
@djberg96 @bzwei do you want to have a look at that? 
The dialog.rb and miq_dialog.rb do mostly the same.
Otherwise I'll do it.

bzwei:
@durandom Go ahead. Thanks!

djberg96:
Done!

PR #58, opened 2017-04-25
Move service dialog YAML file to provider repo

djberg96:
This moves the azure-single-vm-from-user-image.yml file from the core repo to this provider repo.

As requested by @durandom at https://github.com/ManageIQ/manageiq-providers-azure/issues/46


djberg96:
Should be done in conjunction with https://github.com/ManageIQ/manageiq/pull/14873

PR #62, opened 2017-04-27
[REFACTOR] Set instance location to region location instead of using uid value

djberg96:
I'm not sure why we're setting the `location` for a vm instance to the `uid` value. This PR sets it to the location (region) value, e.g. "eastus".


jerryk55:
miq-bot add_label fine/yes

djberg96:
@jerryk55 Another possible solution is to access ems.provider region if you have access to it.

jerryk55:
@djberg96 that is true but that will require modifying an existing PRs rather than just back-porting this PR.  Your mileage may vary but I would think this is the preferred route.

simaishi:
Fine backport details:

```
$ git log -1
commit b75fb9b42252ac693710b7314b7e0d99462943fb
Date:   Fri Apr 28 16:02:04 2017 -0400

    Merge pull request #62 from djberg96/instance_location
    
    [REFACTOR] Set instance location to region location instead of using uid value
    (cherry picked from commit 04e380b10436073bbc291dc75124bd5e031e5f0d)
    
```

PR #65, opened 2017-05-04
Add support for managed images

djberg96:
This PR adds support for inventory and provisioning of managed images.

Addresses https://bugzilla.redhat.com/show_bug.cgi?id=1439815


djberg96:
@Ladas @bronaghs Please review.

Note that at the moment I'm using the `ems_ref` to distinguish between managed and unmanaged. This isn't really ideal, but I think the ideal implementation has us implementing a StorageManager.

So, the plan is to use this for now, and refactor that once we have a more robust way of checking the storage type.

app/models/manageiq/providers/azure/cloud_manager/refresh_parser.rb
```diff
@@ -477,6 +484,35 @@ def parse_stack_resource(resource, group)
         return uid, new_result
       end
 
+      def parse_managed_image(image)
+        uid = image.id.downcase
+
+        resource_group = ResourceGroup.where(:name => image.resource_group.downcase).find do |rg|
```
djberg96: Which is fine, we'll just get nil then.

bronaghs: Why was it in there in the first place?

app/models/manageiq/providers/azure/cloud_manager/provision/cloning.rb
```diff
@@ -68,17 +74,25 @@ def prepare_for_clone_task
...
+    if target_uri
+      cloud_options[:properties][:storageProfile][:osDisk][:name]  = dest_name + SecureRandom.uuid + '.vhd'
+      cloud_options[:properties][:storageProfile][:osDisk][:image] = {:uri => source_uri}
+      cloud_options[:properties][:storageProfile][:osDisk][:vhd]   = {:uri => target_uri}
+    else
+      # Default to a storage account type of "Standard_LRS" for managed images for now.
+      cloud_options[:properties][:storageProfile][:osDisk][:managedDisk] = {:storageAccountType => 'Standard_LRS'}
+      cloud_options[:properties][:storageProfile][:imageReference] = {:id => source.ems_ref}
+    end
+
```
djberg96: No, with managed disks, Azure handles that for us.

app/models/manageiq/providers/azure/cloud_manager/refresh_parser.rb
djberg96: Because I would like to set that relationship in the database for the sake of ActiveRecord, if possible. Then we aren't parsing resource group information out of strings, but instead are using Rails methods.

Ladas: Ah, @bronaghs  is right, we should use @data_index and then make sure this relations is stored in save_inventory.

Also this creates n+1 queries, so if @data_index is not enough, we should use different hash  and then do O(1) lookup. We can get rid of the include? right?

djberg96: @bronaghs I think I did that because I thought I might need it in the future for provisioning. But at the moment, no, we don't use it explicitly anywhere.

app/models/manageiq/providers/azure/cloud_manager/refresh_parser.rb
```diff
@@ -477,6 +484,35 @@ def parse_stack_resource(resource, group)
...
+          :location           => @ems.provider_region,
+          :vendor             => 'azure',
+          :raw_power_state    => 'never',
+          :template           => true,
+          :publicly_available => false,
+          :resource_group_id  => resource_group.try(:id),
+          :operating_system   => process_os(image),
+          :hardware           => {
+            :bitness  => 64,
+            :guest_os => image.properties.storage_profile.os_disk.os_type
```
Ladas: I wonder if we could use few try() here, can it be the specific properties are not always there?

djberg96: I've not seen one yet that was, but I can't rule out the possibility. What should it default to, if anything?

Ladas: I see ``` || "unknown" ``` elsewhere

Ladas: hm, you should probably use the normalize_os method we have? https://github.com/Ladas/manageiq-providers-amazon/blob/f8615fb4c69e7485b57f4f1ffd4f6bbbc4782438/app/models/manageiq/providers/amazon/inventory/parser/cloud_manager.rb#L73-L73

Ladas: also amazon just puts the "linux" if it's "unknown" :-)

bronaghs: Really? Thats a bit assumption.

djberg96: @Ladas You realize that method will bomb if it receives a nil argument, right?

djberg96:
@bronaghs @Ladas Ok, I've removed the resource group stuff for now, and add a couple "try" statements for the OS.

djberg96:
@miq-bot add_label fine/yes

djberg96:
Actually, we need to backport https://github.com/ManageIQ/manageiq-providers-azure/pull/51 in order to backport to Fine.

app/models/manageiq/providers/azure/cloud_manager/refresh_parser.rb
```diff
@@ -477,6 +484,30 @@ def parse_stack_resource(resource, group)
...
+          :description        => "#{image.resource_group}\\#{image.name}",
+          :location           => @ems.provider_region,
+          :vendor             => 'azure',
+          :raw_power_state    => 'never',
+          :template           => true,
+          :publicly_available => false,
+          :operating_system   => process_os(image),
+          :hardware           => {
+            :bitness  => 64,
+            :guest_os => image.properties.storage_profile.try(:os_disk).try(:os_type) || 'unknown'
```
Ladas: right, didn't try the nil sent to normalize method, but the normalize method should be used, since we do infer the correct OS icons to show in the UI from this field.

djberg96:
@Ladas Ok, I've updated the OS name handling.

app/models/manageiq/providers/azure/cloud_manager/refresh_parser.rb
```diff
@@ -477,6 +484,32 @@ def parse_stack_resource(resource, group)
...
+          :location           => @ems.provider_region,
+          :vendor             => 'azure',
+          :raw_power_state    => 'never',
+          :template           => true,
+          :publicly_available => false,
+          :operating_system   => process_os(image),
+          :hardware           => {
+            :bitness  => 64,
+            :guest_os => OperatingSystem.normalize_os_name(os)
+          }
```
bronaghs: @djberg96 - Why is there both an `operating_system` and `hardware[:guest_os]` property?

djberg96: @bronaghs I'm just aping the existing code. My assumption was that it creates DB relationships properly so that you can do `vm.operating_system` or `vm.hardware.guest_os`.

app/models/manageiq/providers/azure/cloud_manager/provision/cloning.rb
bronaghs: OK

app/models/manageiq/providers/azure/cloud_manager/refresh_parser.rb
bronaghs: OK

Ladas:
looks great 👍

simaishi:
Backported to Euwe via https://github.com/ManageIQ/manageiq/pull/15164

djberg96:
@simaishi Still need this backported to Fine please.

simaishi:
@djberg96 I assume you want to keep `~>` in .gemspec?

```
s.add_dependency "azure-armrest", "~>0.7.2"
```
https://github.com/ManageIQ/manageiq-providers-azure/pull/55 is `fine/no`, so I just want to make sure.

djberg96:
@simaishi Yes, let's keep that.

simaishi:
Fine backport details:

```
$ git log -1
commit 5e1fb1c45b5671b7c891658cf8e19ef28740e83c
Date:   Wed May 10 15:02:39 2017 -0400

    Merge pull request #65 from djberg96/managed_images2
    
    Add support for managed images
    (cherry picked from commit 3212ee2ae4a477426434325e3704e003245de89b)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1459217
```

Issue #69, opened 2017-05-17
Refresh parser should set resource group where possible


djberg96:
Currently the refresh parser isn't setting the resource group relationship for any resources (except stacks, which uses a string).

We should be setting the relationship so that we can get that information internally via ActiveRecord when needed.

PR #72, opened 2017-05-22
Create resource group association for instances and managed images

djberg96:
At the moment we collect resource group information, but don't actually create the appropriate association between a resource and its resource group. This PR attempts to implement that, starting with VM instances and managed images.

Note that this will require some modifications to the core repo before it actually works properly. Without them this does nothing of note, though it doesn't cause any breakage.

Addresses https://github.com/ManageIQ/manageiq-providers-azure/issues/69

Working in conjunction with https://github.com/ManageIQ/manageiq/pull/15187

### DIFF
--- a/app/models/manageiq/providers/azure/cloud_manager/refresh_parser.rb
+++ b/app/models/manageiq/providers/azure/cloud_manager/refresh_parser.rb
@@ -229,6 +229,8 @@ module ManageIQ::Providers
         # unified under the NetworkManager
         hardware_network_info = get_hardware_network_info(instance)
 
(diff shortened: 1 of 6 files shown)


djberg96:
@miq-bot add_label wip

djberg96:
Note that this PR will get type mismatch errors until the change in the core repo is made.

djberg96:
The core repo was updated:

https://github.com/ManageIQ/manageiq/pull/14000
https://github.com/ManageIQ/manageiq/pull/14948
https://github.com/ManageIQ/manageiq/pull/15187

PR #81, opened 2017-06-22
Disable reset operation for VMs and add specs

djberg96:
Explicitly remove support for hard resets for Azure.

This partially addresses https://github.com/ManageIQ/manageiq/issues/9684. It will also require a change on the classic UI side, which is currently hard-coding a value for the instances view that is keeping the option around.

On a side note, I reorganized the specs a bit so that some of the variables were available to other contexts.


simaishi:
@djberg96 I'm removing `fine/yes` label, as it looks this change was never completed. If it's still needed in Fine, please add BZ link as well as related PRs. Thanks!

PR #86, opened 2017-06-27
Decrypt client keys in raw connect

jntullo:
Working on switching credential validation onto the queue using the class method raw_connect to be able to validate in different zones and remove the use of a temporary EMS. This will require passwords to be encrypted / decrypted in raw_connect.

More info: https://github.com/ManageIQ/manageiq-ui-classic/pull/1580

### DIFF
--- a/app/models/manageiq/providers/azure/manager_mixin.rb
+++ b/app/models/manageiq/providers/azure/manager_mixin.rb
@@ -38,7 +38,7 @@ module ManageIQ::Providers::Azure::ManagerMixin
 
       ::Azure::Armrest::Configuration.new(
         :client_id       => client_id,
-        :client_key      => client_key,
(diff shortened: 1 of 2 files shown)


djberg96:
@jntullo If the password isn't encrypted currently, what happens?

jntullo:
@djberg96 `.try_decrypt` checks if it's encrypted. So if you pass in an argument that isn't encrypted, it will still be fine.

djberg96:
I don't *entirely* understand the purpose of this, but it seems ok to me. I will let @blomquisg push the merge button, though.

blomquisg:
This seems fine since the password its decrypting is passed in by the caller and not coming out the DB. 
 And, since it still handled unencrypted passwords for connection validation.

PR #89, opened 2017-07-07
Add ems_azure_with_vcr_authentication factory

djberg96:
This is a minor refactoring for rspec that abstracts some of our authentication setup into its own factory. This way we won't have to repeat this pattern in setup methods for rspec tests.

It is modeled after similar Amazon code:

https://github.com/ManageIQ/manageiq-providers-amazon/blob/master/spec/factories/ext_management_system.rb


bronaghs:
@djberg96 - are you going to use this anywhere?

djberg96:
@bronaghs yes, i want to refactor the existing refresher spec, for example. Should be useful for other future specs, too.

PR #95, opened 2017-07-17
Marketplace image support

01100010011001010110010101110000:
This allows ```refresh_parser``` to fetch marketplace images, filtered by URNs supplied in settings.yml, and handles provisioning of instances from said images by matching against the format of a marketplace URN in ```prepare_for_clone_task```

### DIFF
--- a/app/models/manageiq/providers/azure/cloud_manager/provision/cloning.rb
+++ b/app/models/manageiq/providers/azure/cloud_manager/provision/cloning.rb
@@ -51,7 +51,7 @@ module ManageIQ::Providers::Azure::CloudManager::Provision::Cloning
 
     # TODO: Ideally this would be a check against source.storage or source.disks
     if source.ems_ref =~ /.+:.+:.+:.+/
-      urn_keys = ['publisher', 'offer', 'sku', 'version']
+      urn_keys = %w[publisher offer sku version]
       image_reference = Hash[urn_keys.zip(source.ems_ref.split(':'))]
       os, target_uri, source_uri = nil
     elsif source.ems_ref.starts_with?('/subscriptions')
--- a/app/models/manageiq/providers/azure/cloud_manager/provision/cloning.rb
+++ b/app/models/manageiq/providers/azure/cloud_manager/provision/cloning.rb
@@ -50,10 +50,16 @@ module ManageIQ::Providers::Azure::CloudManager::Provision::Cloning
     nic_id = associated_nic || create_nic
 
     # TODO: Ideally this would be a check against source.storage or source.disks
-    if source.ems_ref.starts_with?('/subscriptions')
+    if source.ems_ref =~ /.+:.+:.+:.+/
+      urn_keys = ['publisher', 'offer', 'sku', 'version']
+      image_reference = Hash[urn_keys.zip(source.ems_ref.split(':'))]
+      os, target_uri, source_uri = nil
+    elsif source.ems_ref.starts_with?('/subscriptions')
       os = source.operating_system.product_name
       target_uri, source_uri = nil
+      image_reference = { :id => source.ems_ref }
     else
+      image_reference = nil
       target_uri, source_uri, os = gather_storage_account_properties
     end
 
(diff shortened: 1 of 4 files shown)


app/models/manageiq/providers/azure/cloud_manager/refresh_parser.rb
```diff
@@ -24,6 +24,7 @@ def initialize(ems, options = nil)
         @rgs               = resource_group_service(@config)
         @sas               = storage_account_service(@config)
         @mis               = managed_image_service(@config)
+        @vmis              = virtual_machine_image_service(@config, :location => @ems.provider_region)
```
djberg96: I don't think setting `:location` here actually does anything.

app/models/manageiq/providers/azure/cloud_manager/provision/cloning.rb
```diff
@@ -50,10 +50,16 @@ def prepare_for_clone_task
     nic_id = associated_nic || create_nic
 
     # TODO: Ideally this would be a check against source.storage or source.disks
-    if source.ems_ref.starts_with?('/subscriptions')
+    if source.ems_ref =~ /.+:.+:.+:.+/
```
djberg96: Does it need to be this tricky? It could just be `.starts_with("http")?` couldn't it?

app/models/manageiq/providers/azure/cloud_manager/refresh_parser.rb
01100010011001010110010101110000: I believe it sets the location used by that instance of vmis, which I had to do because ```gather_data_for_this_region``` filters by location, and from what I could tell there wasn't a way to pass arguments to the ```list_all``` call that takes place inside ```gather_data_for_this_region```

djberg96: Ah, right, I forgot the constructor for VirtualMachineImageService is a little different and accepts a location argument.

01100010011001010110010101110000:
By the way, what're the odds on getting this backported to fine?

djberg96:
@blomquisg Assuming this is merged, any reason we can't backport to Fine? We just need a BZ, right?

app/models/manageiq/providers/azure/cloud_manager/refresh_parser.rb
```diff
@@ -167,6 +169,18 @@ def get_managed_images
...
 
+      def get_market_images
+        urns   = @options.market_image_urns
+        images = if urns
+                   gather_data_for_this_region(@vmis).select do |image|
+                     urns.include?(image.id)
+                   end
+                 else
+                   gather_data_for_this_region(@vmis)
+                 end
```
01100010011001010110010101110000: That is correct; the last time I ran this command, it gathered approximately 4500 images. I'm by no means certain if that's a reasonable default. If they just flip ```@options.market_image_urns``` to true, there are some URNs already defined to avoid that.

bronaghs: @01100010011001010110010101110000 - yea I realize that we can filter on URN, I'm just concerned we could end up gathering a ton of images, the vast majority of which would never be used in provisioning, and it could become a performance issue.

@blomquisg - Do you think filtering the gathering of marketplace images by URN is enough to address a possible performance issue?

blomquisg: We do the same thing for Amazon, honestly.  If the user enables public images and does not set a filter on which public images to pull in, they're opening themselves up for 10k+ images.

And, yeah, it has a performance impact in inventory collection, especially since we gather them each time we do inventory collection to do the inventory comparisons.

There is work underway in inventory collection optimizations that actually improve this situation quite a bit (for those that actually want to see *all* those images).  But, that's a ways off.

I'd say for now, since we do it already for another provider (and arguably, more than double the images), it's probably OK here.

Maybe just add a simple comment to the settings.yml file indicating that neglecting to set a filter and enabling marketplace images will result in much slower inventory collection times, and it's advisable to set a marketplace image filter.

djberg96: `az vm image list --output json --all` seems to be considerably faster than our custom method. I'm not sure why yet, still digging through the output (with --debug attached to the command).

Maybe we can speed things up and/or just cache the results.

bronaghs:
@01100010011001010110010101110000 - if you could address the rubocop complaints I will merge this once the tests pass and once https://github.com/ManageIQ/manageiq-providers-azure/pull/97 is merged. (you can ignore the complaint around the method name).

01100010011001010110010101110000:
I believe that should handle it. Is there some rubocop wrapper to inspect only the contents of a git diff? Couldn't find anything in a cursory google

djberg96:
👍 LGTM for now. I do think we may need to do some refactoring to speed things up, but I also believe that burden lies mostly on the azure-armrest gem.

PR #97, opened 2017-07-20
Upgrade azure-armrest gem to 0.8.0

djberg96:
This updates the azure-armrest gem to 0.8.0. The major version bump was because some return types were changed in the Storage model for a couple methods. However, the methods that were altered are not currently used in ManageIQ.

In addition to other updates, this contains the `VirtualMachineImage#list_all` method, which we need for marketplace image support: https://github.com/ManageIQ/manageiq-providers-azure/pull/95

### DIFF
--- a/manageiq-providers-azure.gemspec
+++ b/manageiq-providers-azure.gemspec
@@ -13,7 +13,7 @@ Gem::Specification.new do |s|
 
   s.files = Dir["{app,config,lib}/**/*"]
 
-  s.add_dependency "azure-armrest", "~>0.7.3"
+  s.add_dependency "azure-armrest", "~>0.8.0"
 
   s.add_development_dependency "codeclimate-test-reporter", "~> 1.0.0"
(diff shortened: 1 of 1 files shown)


bronaghs:
@Fryguy - In the past we would point to a specific version of any gem developed by our organization. Recently, I have seen otherwise. Can you clarify if that is still considered a best practice? Or do we not care anymore?

Fryguy:
@bronaghs I don't understand what you mean.  This PR is correct.

bronaghs:
@Fryguy - A couple of years ago we released a new version of the azure-armrest gem but had not updated the provider code in the manageiq repo to be compatible with it, automatically pulling in the new gem version caused travis failures in the manageiq repo. At that time I was told we should always point to a specific version of any gems we wrote ourselves. Looking at our gemspecs, I am possibly the only person adhering to that, so I was asking for clarification from you on whether this is something I need to be concerned about - by merging this PR you have clarified that I don't.

PR #98, opened 2017-07-21
Add support for regions in Germany

djberg96:
This PR adds support for regions in Germany.

~~~This is a WIP for now until I can get some credentials that actually have access to these regions. There's also the issue of listing regions that may not be available for the given subscription, though the validate button should reveal any issues.~~~

Update: Support for regions in China will be added later in a separate PR.

### DIFF
--- a/app/models/manageiq/providers/azure/manager_mixin.rb
+++ b/app/models/manageiq/providers/azure/manager_mixin.rb
@@ -48,6 +48,8 @@ module ManageIQ::Providers::Azure::ManagerMixin
 
     def environment_for(region)
       case region
+      when /germany/i
+        ::Azure::Armrest::Environment::Germany
       when /usgov/i
         ::Azure::Armrest::Environment::USGovernment
       else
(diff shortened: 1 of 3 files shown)


djberg96:
Confirmed that validation will fail if the subscription does not support the region. The "Add" button is greyed out in the UI until validation is successful.

djberg96:
@miq-bot remove_label wip

djberg96:
I managed to get a service principal setup for Germany and tested it with azure-armrest successfully, so I'm removing the WIP label.

bronaghs:
If you only tested with Germany, the China regions should be added later.

bronaghs:
hey @djberg96 - What is the status of this PR?
I saw an email notification stating that the regions support has since changed on the Azure side, would be worthwhile checking to see if we need to update our supported list.

djberg96:
@bronaghs I was hoping to get some credentials for a China region before messing with this PR, but that didn't happen. I'll double check the current list and see if there are any other regions to add.

djberg96:
@bronaghs Ok, I've removed references to China for now.

djberg96:
@bronaghs Updated.

bronaghs:
@djberg96 - Any reason why support for Germany cannot be backported to fine?

djberg96:
@bronaghs As long as there's a BZ, should be ok. :)

PR #103, opened 2017-08-02
Add additional_regions support, and add specs for regions

djberg96:
I noticed that the Amazon provider code allowed for an extra field in the settings called "additional_regions". I decided that this could be useful for us as well. If another region is added but we haven't updated regions.rb yet, the users could add it to their settings file. Granted, it doesn't deal with the possibility of different endpoints, but it's still nice to have.

I also noticed that we had no specs for regions, so I've added them as well. They are mosty modeled on the Amazon specs.

On a side note, I'm not sure why I have to explicitly require azure-armrest, but if I don't put it there I get an undefined constant error. Suggestions welcome.


jteehan:
Is this in the same section as disabled regions?  I assume this will allow customers to add new regions between releases?

bronaghs:
Thats right @jteehan

bronaghs:
@djberg96 - Travis failure looks legit

djberg96:
@bronaghs I can't duplicate that error locally. It's like it's not using the .yml file on Travis for some reason.

PR #104, opened 2017-08-02
Refactor discovery code, eliminate a warning

djberg96:
This fixes a warning that was introduced in azure-armrest 0.8.0 by switching from the `locations` method to the `list_locations` method.

More significantly, it refactors the discovery code to use the `ResourceService` class to get a count of VM's. Since we're only checking for the count of VM's in the discovery block and don't care about the details, it's better to use the `ResourceService#list_all` method with a filter.

This is because the Azure REST API allows filter parameters up-front on the query, unlike our current method which actually filters after the fact. In addition, much less JSON is returned when you use the resource API instead of our current code, which saves memory.

For benchmarking I did this:

```
require 'azure-armrest'
require 'benchmark'

conf = Azure::Armrest::Configuration.new(cred)

res = Azure::Armrest::ResourceService.new(conf)
vms = Azure::Armrest::VirtualMachineService.new(conf)

filter = "resourceType eq 'Microsoft.Compute/virtualMachines' and location eq 'eastus'"

Benchmark.bm(30) do |x|
  x.report("resource"){
    3.times{ res.list_all(:all => true, :filter => filter) }
  }
  x.report("list_all"){
    3.times{ vms.list_all(:location => "eastus") }
  }
end
```

The results were:
```
                                     user     system      total        real
resource                         0.130000   0.010000   0.140000 (  6.400723)
list_all                         1.400000   0.020000   1.420000 ( 11.854761)
```

So, a 2x performance improvement in terms of real time!

For memory, I just did a simple rss check by using
```
pid, size = `ps ax -o pid,rss | grep -E "^[[:space:]]*#{$$}"`.strip.split.map(&:to_i)
puts "PID: #{pid}, SIZE: #{size}"
```
What I saw was about a 40-50% reduction in memory when I used the resource approach.


bronaghs:
Great improvement @djberg96

PR #114, opened 2017-08-18
Updated metric names for Azure metrics

bronaghs:
It appears that the Azure metrics names for CPU, Disk, Memory and Network have slightly changed for most, but not all, VMs eg:
`\\Processor(_Total)\\% Processor Time` has evolved to `\\Processor Information(_Total)\\% Processor Time`

This PR adds these new metric names while keeping the original version that some VMs seem to still respond to.

I am going to keep this a WIP while I finish writing specs.

https://bugzilla.redhat.com/show_bug.cgi?id=1476756

### DIFF
--- a/app/models/manageiq/providers/azure/cloud_manager/metrics_capture.rb
+++ b/app/models/manageiq/providers/azure/cloud_manager/metrics_capture.rb
@@ -2,12 +2,21 @@ class ManageIQ::Providers::Azure::CloudManager::MetricsCapture < ManageIQ::Provi
   INTERVAL_1_MINUTE = "PT1M".freeze
 
   # Linux, Windows counters
-  CPU_METERS     = ["\\Processor(_Total)\\% Processor Time", "\\Processor\\PercentProcessorTime"].freeze
+  CPU_METERS     = ["\\Processor Information(_Total)\\% Processor Time",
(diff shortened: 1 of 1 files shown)


djberg96:
@bronaghs Does it change depending on the api-version string? I wouldn't think so since we lock down the api-version strings now, but it's worth a test.

There's another option, to, which is to use the "universal" metrics that are supported across all platforms. More info at https://github.com/ManageIQ/azure-armrest/wiki/Insights:-Metrics

bronaghs:
@djberg96 - Although I see the value in universal metrics, the platform is not the problem there. The problem is that new counters have been added and new VMs deployed since a specific date are configured to use those new counters. So it doesn't sound like the universal metrics will help in this particular case, is that correct?

simaishi:
Fine backport details:

```
$ git log -1
commit 13d57decad9a1067d5f25e961f2f0f78b793c828
Date:   Mon Aug 21 10:07:58 2017 -0600

    Merge pull request #114 from bronaghs/new_metric_names_bz_1476756
    
    Updated metric names for Azure metrics
    (cherry picked from commit 09581c9275e5332eafa0901640705e36f4f57886)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1484374
```

PR #115, opened 2017-08-21
Refer to resource group object, not string

bronaghs:
The resource group property on a VM was previously a string. Recently, this was changed to a ResourceGroup object.
Refer now to `target.resource_group.name` in the metrics capture code.

https://bugzilla.redhat.com/show_bug.cgi?id=1476756


djberg96:
The change in question happened with https://github.com/ManageIQ/manageiq-providers-azure/pull/72.

Note that PR (and the underlying ManageIQ PR's) were not backported to Fine or Euwe since this was just an enhancement. Consequently, you'll probably want to put a guard in like so:

`resource_group = target.resource_group.try(:name) || target.resource_group`

bronaghs:
Good tip - thanks

bronaghs:
@djberg96 - yea I havent pushed it up yet, got distracted :-)

bronaghs:
@djberg96 - this has the fine/no label set, so I dont think the guard is needed. No?

PR #116, opened 2017-08-21
Update for sprint 67

JPrause:


bronaghs:
@JPrause - Looks like you missed: https://github.com/ManageIQ/manageiq-providers-azure/pull/114

PR #117, opened 2017-08-24
Add Snapshot Code for Azure Managed Disks

jerryk55:
Add code to snapshot VMs with Managed Disks,
as well as the code to delete the snapshots upon
completion of SSA (or aborts).
Also check that the disk requires a snapshot, which for now
is just for Managed Disks.  This will change in a subsequent PR.

This is one of several PRs required to add support for SSA for Managed Disks,
that is is response to the following BZs:
https://bugzilla.redhat.com/show_bug.cgi?id=1475540
and
https://bugzilla.redhat.com/show_bug.cgi?id=1459612

A PR for the azure-armrest gem has already been merged in advance of this:
https://github.com/ManageIQ/azure-armrest#299

A PR for the ManageIQ repo has been added as well to work with this -
https://github.com/ManageIQ/manageiq#15865

Also a PR for the manageiq-smartstate gem is required for this - 
https://github.com/ManageIQ/manageiq-smartstate/pull/23

**# Links**

https://bugzilla.redhat.com/show_bug.cgi?id=1475540
https://bugzilla.redhat.com/show_bug.cgi?id=1459612
ManageIQ/azure-armrest#299
ManageIQ/manageiq#15865
ManageIQ/manageiq-smartstate#23

**# Steps for Testing/QA**

Add an Azure VM with a Managed Disk.
Run SmartState Analysis against the VM.

@roliveri @hsong-rh please review for SSA sanity.
@bronaghs @blomquisg please review or assign an appropriate providers person to do so.
Thanks!

### DIFF
--- a/app/models/manageiq/providers/azure/cloud_manager.rb
+++ b/app/models/manageiq/providers/azure/cloud_manager.rb
@@ -126,9 +126,11 @@ class ManageIQ::Providers::Azure::CloudManager < ManageIQ::Providers::CloudManag
                      :properties => {
                        :creationData => {
                          :createOption     => "Copy",
-                         :sourceResourceId => os_disk.managed_disk.id } } }
+                         :sourceResourceId => os_disk.managed_disk.id
+                       }
+                     } }
     snap_name = os_disk.name + "__EVM__SSA__SNAPSHOT"
-    _log.debug ("vm=[#{vm.name}] creating SSA snapshot #{snap_name}")
+    _log.debug("vm=[#{vm.name}] creating SSA snapshot #{snap_name}")
     begin
       snap_svc.get(snap_name, vm.resource_group)
     rescue ::Azure::Armrest::NotFoundException, ::Azure::Armrest::ResourceNotFoundException => err
@@ -136,12 +138,12 @@ class ManageIQ::Providers::Azure::CloudManager < ManageIQ::Providers::CloudManag
         snap_svc.create(snap_name, vm.resource_group, snap_options)
         return snap_name
       rescue => err
-        _log.error ("vm=[#{vm.name}], error: #{err}")
+        _log.error("vm=[#{vm.name}], error: #{err}")
         _log.debug { err.backtrace.join("\n") }
         raise "Error #{err} creating SSA Snapshot #{snap_name}"
       end
     end
-    _log.error ("SSA Snapshot #{snap_name} already exists.")
+    _log.error("SSA Snapshot #{snap_name} already exists.")
     raise "Snapshot #{snap_name} already exists. Another SSA request for this VM is in progress or a previous one failed to clean up properly."
   end
 
@@ -152,15 +154,15 @@ class ManageIQ::Providers::Azure::CloudManager < ManageIQ::Providers::CloudManag
     vm_obj    = vm_svc.get(vm.name, vm.resource_group)
     os_disk   = vm_obj.properties.storage_profile.os_disk
     snap_name = os_disk.name + "__EVM__SSA__SNAPSHOT"
-    _log.debug ("vm=[#{vm.name}] deleting SSA snapshot #{snap_name}")
+    _log.debug("vm=[#{vm.name}] deleting SSA snapshot #{snap_name}")
     snap_svc.delete(snap_name, vm.resource_group)
   rescue => err
-    _log.error ("vm=[#{vm.name}], error: #{err}")
+    _log.error("vm=[#{vm.name}], error: #{err}")
     _log.debug { err.backtrace.join("\n") }
   end
 
   def snapshot_service(connection = nil)
-    _log.debug ("Enter")
+    _log.debug("Enter")
     connection ||= connect
     ::Azure::Armrest::Storage::SnapshotService.new(connection)
   end
--- a/app/models/manageiq/providers/azure/cloud_manager.rb
+++ b/app/models/manageiq/providers/azure/cloud_manager.rb
@@ -34,6 +34,8 @@ class ManageIQ::Providers::Azure::CloudManager < ManageIQ::Providers::CloudManag
   before_create :ensure_managers
   before_update :ensure_managers_zone_and_provider_region
 
+  SSA_SNAPSHOT_SUFFIX = "__EVM__SSA__SNAPSHOT".freeze
+
   # If the Microsoft.Insights Azure provider is not registered, then neither
   # events nor metrics are supported for that EMS.
   #
@@ -129,7 +131,7 @@ class ManageIQ::Providers::Azure::CloudManager < ManageIQ::Providers::CloudManag
                          :sourceResourceId => os_disk.managed_disk.id
                        }
                      } }
-    snap_name = os_disk.name + "__EVM__SSA__SNAPSHOT"
+    snap_name = "#{os_disk.name}#{SSA_SNAPSHOT_SUFFIX}"
     _log.debug("vm=[#{vm.name}] creating SSA snapshot #{snap_name}")
     begin
       snap_svc.get(snap_name, vm.resource_group)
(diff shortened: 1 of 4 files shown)


app/models/manageiq/providers/azure/cloud_manager/vm_or_template_shared/scanning.rb
```diff
@@ -56,4 +56,18 @@ def has_proxy?
   def requires_storage_for_scan?
     false
   end
+
+  def require_snapshot_for_scan?
```
roliveri: @jerryk55 So, when you add snapshots for non-managed disks, this method will always return true?

jerryk55: @roliveri no, when you check to see if you need to add a snapshot for non-managed disks, it will always return false, won't it?

roliveri: @jerryk55 I thought you said that they suggested that both, managed and non-managed disks, always be snapshotted.

jerryk55: Oh yes I misunderstood your original question.  Yes - when I add the other code for non-managed disk snapshots - this method will simply return true all the time.

jerryk55:
@bronaghs @blomquisg @djberg96 Can someone on the Providers team review and merge please?  We have outstanding BZs waiting for this as well as other PRs that have been merged waiting for this set of changes.  Thanks.

bdunne:
@jerryk55 Are you going to address the rubocop comments first?

jerryk55:
@bdunne - ah, the RuboCop comments.  The first go-round RuboCop complained because there weren't parentheses around the arguments to the log message calls.  So I put them in.  And now its complaining because there ARE parentheses around the log message calls.   So I would like to throw Rubocop in the garbage, thank you very much.

jerryk55:
@bdunne excuse my rant.  :-)  I've been absolved of all warnings by the almighty Rubocop.

app/models/manageiq/providers/azure/cloud_manager.rb
```diff
@@ -114,4 +114,54 @@ def vm_reboot_guest(vm, _options = {})
...
+        snap_svc.create(snap_name, vm.resource_group, snap_options)
+        return snap_name
+      rescue => err
+        _log.error ("vm=[#{vm.name}], error: #{err}")
+        _log.debug { err.backtrace.join("\n") }
+        raise "Error #{err} creating SSA Snapshot #{snap_name}"
+      end
+    end
+    _log.error ("SSA Snapshot #{snap_name} already exists.")
+    raise "Snapshot #{snap_name} already exists. Another SSA request for this VM is in progress or a previous one failed to clean up properly."
```
blomquisg: Who calls this method?  I'm seeing the possibility of a caller having to react to two different error conditions in possibly two different ways.  One error condition is when it can't create the snapshot because of an error (probably in Azure).  And another error condition when an SSA request is already being processed for this VM.

Maybe the "snapshot already exists" situation could be a specific error type?  Then, the caller could rescue that error type in its own block with a special handler.  And all other errors could be assumed to be an actual failure to process the request by the provider.

I could also see the "snapshot already exists" situation being a non-error, and just log a warning and return `nil`.  But, that might be too ambiguous for the caller.

Thoughts?

app/models/manageiq/providers/azure/cloud_manager.rb
```diff
@@ -114,4 +114,56 @@ def vm_reboot_guest(vm, _options = {})
...
+        snap_svc.create(snap_name, vm.resource_group, snap_options)
+        return snap_name
+      rescue => err
+        _log.error("vm=[#{vm.name}], error: #{err}")
+        _log.debug { err.backtrace.join("\n") }
+        raise "Error #{err} creating SSA Snapshot #{snap_name}"
+      end
+    end
+    _log.error("SSA Snapshot #{snap_name} already exists.")
+    raise "Snapshot #{snap_name} already exists. Another SSA request for this VM is in progress or a previous one failed to clean up properly."
```
blomquisg: Who calls this method? I'm seeing the possibility of a caller having to react to two different error conditions in possibly two different ways. One error condition is when it can't create the snapshot because of an error (probably in Azure). And another error condition when an SSA request is already being processed for this VM.

Maybe the "snapshot already exists" situation could be a specific error type? Then, the caller could rescue that error type in its own block with a special handler. And all other errors could be assumed to be an actual failure to process the request by the provider.

I could also see the "snapshot already exists" situation being a non-error, and just log a warning and return nil. But, that might be too ambiguous for the caller.

Thoughts?

jerryk55: @blomquisg It is called by VmScan.create_snapshot. We have similar logic for the Scvmm code as well (although some of it is embedded in the manageiq-smartstate repo. 

It really doesn't matter to VmScan if we can't perform the snapshot because of a provider error or because the snapshot already exists - the SSA request needs to fail in either case.

The "snapshot already exists" situation cannot be a non-error - if there really is another request to run SSA on the same instance/image that we somehow missed (which should not happen) a problem would arise when the first request finished and therefore deleted the snapshot.  Since these requests are stateless and the snapshot disk isn't actually mounted or opened the second request would then run into an unexplained error.

Let me know what you think.  @roliveri, feel free to chime in if you feel otherwise.

roliveri: Historically, `vm_create_evm_snapshot` is not a general purpose snapshot creation method, `evm` snapshots are only created for SSA. SSA doesn't allow multiple `evm` (SSA) snapshots, hence the error.

Maybe there should be a general `create_snapshot` method, that is called by `vm_create_evm_snapshot`.

jerryk55: @roliveri you mean as a follow-on task for *all* the providers needing it, right?

roliveri: Yes. If a provider needs a general create snapshot facility, then it makes sense to structure the code that way. As it stands now, `vm_create_evm_snapshot` is SSA specific, so it should be ok the way it is.

jerryk55:
@blomquisg so please let me know if there are any other questions or if we can get this merged.  Thanks!

app/models/manageiq/providers/azure/cloud_manager.rb
```diff
@@ -114,4 +114,56 @@ def vm_reboot_guest(vm, _options = {})
...
+    return unless vm_obj.managed_disk?
+    os_disk      = vm_obj.properties.storage_profile.os_disk
+    snap_options = { :location   => vm.location,
+                     :properties => {
+                       :creationData => {
+                         :createOption     => "Copy",
+                         :sourceResourceId => os_disk.managed_disk.id
+                       }
+                     } }
+    snap_name = os_disk.name + "__EVM__SSA__SNAPSHOT"
```
Fryguy: Prefer interpolation over concatenation.

jerryk55: @Fryguy understood.  Updated.

jerryk55:
@miq-bot add_label fine/yes

simaishi:
Fine backport details:

```
$ git log -1
commit 2dd69ef21edfc14356aca44bcbc6a254052e2d5a
Author: Greg Blomquist <blomquisg@gmail.com>
Date:   Fri Sep 1 10:15:38 2017 -0400

    Merge pull request #117 from jerryk55/managed_disk_snapshot_support
    
    Add Snapshot Code for Azure Managed Disks
    (cherry picked from commit 4492103ec7bfd198e184e144221cb2b398d023b6)
    
```

PR #111, opened 2017-08-15
 Move OrchestrationTemplateAzure to manageiq-providers-azure

bzwei:
Prepare to move OrchestrationTemplate to each provider
  
   This is a step before the migration.

    https://bugzilla.redhat.com/show_bug.cgi?id=1417021

PR #121, opened 2017-09-07
Refresher sets proper OrchestrationTemplate type

bzwei:
This the continuation of #111. The refresher needs to set the type to the new Azure template class.

### DIFF
--- a/app/models/manageiq/providers/azure/cloud_manager/orchestration_template.rb
+++ b/app/models/manageiq/providers/azure/cloud_manager/orchestration_template.rb
@@ -1,4 +1,8 @@
 class ManageIQ::Providers::Azure::CloudManager::OrchestrationTemplate < ::OrchestrationTemplate
+  def format
+    'json'.freeze
+  end
+
   def parameter_groups
     # Azure format does not have the concept of parameter group
     # Place all parameters in one group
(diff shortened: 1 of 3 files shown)


bzwei:
@miq-bot assign @bronaghs 
@miq-bot add_label refactoring

app/models/manageiq/providers/azure/cloud_manager/orchestration_template.rb
```diff
@@ -1,4 +1,8 @@
 class ManageIQ::Providers::Azure::CloudManager::OrchestrationTemplate < ::OrchestrationTemplate
+  def format
+    'json'.freeze
+  end
```
djberg96: What's this for?

bzwei: It is for near future use. All template classes will have this method to tell the format of the content. For Azure only json is supported.

djberg96:
👍 LGTM

app/models/manageiq/providers/azure/cloud_manager/orchestration_template.rb
bronaghs: ah this will pave the way for yaml format soon Im guessing.

PR #122, opened 2017-09-11
Snapshot Blob Disks for SSA

jerryk55:
Add support to Snapshot Blob (non-Managed) disks for SSA.
Previously only Managed Disks were snapshotted.

This PR is one of three in support of a fix to https://bugzilla.redhat.com/show_bug.cgi?id=1463780.
The other two are https://github.com/ManageIQ/manageiq/pull/15960 and https://github.com/ManageIQ/manageiq-smartstate/pull/26

These three PRs should be able to be merged in any order.
@roliveri please review for SSA functionality.  @bronaghs or @djberg96 please review and merge when appropriate.  Thank you.

### DIFF
--- a/app/models/manageiq/providers/azure/cloud_manager.rb
+++ b/app/models/manageiq/providers/azure/cloud_manager.rb
@@ -135,23 +135,23 @@ class ManageIQ::Providers::Azure::CloudManager < ManageIQ::Providers::CloudManag
                          :sourceResourceId => vm.os_disk.managed_disk.id
                        }
                      } }
-    _log.debug("vm=[#{vm.name}] creating SSA snapshot #{vm.snap_name}")
+    _log.debug("vm=[#{vm.name}] creating SSA snapshot #{vm.ssa_snap_name}")
     begin
-      snap_name      = vm.snap_name
+      ssa_snap_name  = vm.ssa_snap_name
       resource_group = vm.resource_group.name
-      snap_svc.get(snap_name, resource_group)
+      snap_svc.get(ssa_snap_name, resource_group)
     rescue ::Azure::Armrest::NotFoundException, ::Azure::Armrest::ResourceNotFoundException => err
       begin
-        snap_svc.create(snap_name, resource_group, snap_options)
-        return snap_name
+        snap_svc.create(ssa_snap_name, resource_group, snap_options)
+        return ssa_snap_name
       rescue => err
         _log.error("vm=[#{vm.name}], error: #{err}")
         _log.debug { err.backtrace.join("\n") }
-        raise "Error #{err} creating SSA Snapshot #{snap_name}"
+        raise "Error #{err} creating SSA Snapshot #{ssa_snap_name}"
       end
     end
-    _log.error("SSA Snapshot #{snap_name} already exists.")
-    raise "Snapshot #{snap_name} already exists. Another SSA request for this VM is in progress or a previous one failed to clean up properly."
+    _log.error("SSA Snapshot #{ssa_snap_name} already exists.")
+    raise "Snapshot #{ssa_snap_name} already exists. Another SSA request for this VM is in progress or a previous one failed to clean up properly."
   end
 
   def vm_create_evm_blob_snapshot(vm)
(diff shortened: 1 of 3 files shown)


app/models/manageiq/providers/azure/cloud_manager/vm_or_template_shared.rb
```diff
@@ -1,9 +1,53 @@
...
+    @vm_object ||= provider_service.get(name, resource_group.name)
+  end
+
+  def os_disk
+    @os_disk ||= vm_object.properties.storage_profile.os_disk
+  end
+
+  delegate :managed_disk?, to: :vm_object
+
+  def snap_name
```
roliveri: Should the name of this method be more specific, like `ssa_snap_name`?

app/models/manageiq/providers/azure/cloud_manager.rb
```diff
@@ -119,26 +117,32 @@ def vm_reboot_guest(vm, _options = {})
...
+    @connection = connect(options)
+    vm.provider_service(@connection)
+    if vm.managed_disk?
+      vm_create_evm_managed_snapshot(vm)
+    else
+      vm_create_evm_blob_snapshot(vm)
+    end
+  end
+
+  def vm_create_evm_managed_snapshot(vm)
```
roliveri: Someday, we should move all of these `vm_*` methods into the their corresponding VM objects (via mixin or whatever is appropriate). This is a general comment that applies to all providers, and shouldn't be addressed in this PR. When this is addressed, we should probably also change the `evm` in names to something like `ssa`.

app/models/manageiq/providers/azure/cloud_manager/vm_or_template_shared.rb
jerryk55: Definitely.  Changing....

app/models/manageiq/providers/azure/cloud_manager.rb
jerryk55: Agreed.  There were suggestions made by @chessbyte on a previous PR in app/models/vm_scan.rb that should probably tackled at the same time.  We should add (an) issue(s) to track this.

app/models/manageiq/providers/azure/cloud_manager/vm_or_template_shared/scanning.rb
```diff
@@ -58,16 +58,6 @@ def requires_storage_for_scan?
...
-        _log.debug("VM #{name} has a Managed Disk - Snapshot required for Scan")
-        return true
-      end
-    rescue => err
-      _log.error("Error Class=#{err.class.name}, Message=#{err.message}")
-    end
-    _log.debug("VM #{name} does not have a Managed Disk - no Snapshot required for Scan")
-    false
+    true
   end
```
bronaghs: @jerryk55 - what is the point of this method if it just returns 'true'

jerryk55:
Bronagh,

Different providers may or may not require
Snapshots for SSA.  The default is false.  Prior to the Managed disk change in the hotfix sent out a few days back, Azure required no snapshots.  With this BZ and its associated PRs Azure always requires a snapshot.  

Jerry

Support My Bike Ride To Cure Type 1 Diabetes

http://www2.jdrf.org/site/TR?fr_id=6889&pg=personal&px=1266643

jerryk55:
@miq-bot add_label fine/yes

djberg96:
@bronaghs @jerryk55 is this a candidate for the SupportsFeatureMixin? I see `:snapshots` in there now, along with a few other snapshot related tags.

simaishi:
Fine backport details:

```
$ git log -1
commit 8d71d0a2895df66ae6d02c0a99eb3c1fb5493a98
Date:   Tue Sep 12 14:02:44 2017 -0400

    Merge pull request #122 from jerryk55/nonmanaged_disk_snapshot_support
    
    Snapshot Blob Disks for SSA
    (cherry picked from commit 29116cc956ae80790e14be488059dc0200770ec6)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1491310
```

PR #123, opened 2017-09-12
Update for Sprint 68

JPrause:
[skip ci]

@miq-bot add_label documentation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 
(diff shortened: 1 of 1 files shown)


bronaghs:
I think one PR was missed: https://github.com/ManageIQ/manageiq-providers-azure/pull/111

JPrause:
When we review PRs for changelog reporting,..we exclude a few things, such as Tests, Technical Debt, Documentation..and in this case Refactoring PRs.

PR111 did show up in my initial report, but was excluded as per ⏫

bronaghs:
I wonder if we should always exclude refactorings,  the PR referred to above is noteworthy. Maybe we can discuss offline for next time.

PR #124, opened 2017-09-13
Don't collect information externally if marketplace images are specified

djberg96:
If users choose to collect marketplace images, and there are already urns specified in the configuration file, then there is no need to use azure-armrest to collect information. We already have all the information we need, since the "id" is just the urn, and everything else can be parsed out of the urn directly.

I've also added some comments explaining that selecting marketplace images without specifying any urns will be expensive and slow.

https://bugzilla.redhat.com/show_bug.cgi?id=1491330

Note that this doesn't address the fundamental issue of `VirtualImageMachineService#list_all` being very slow. I will create a separate issue on the azure-armrest repo.

### DIFF
--- a/app/models/manageiq/providers/azure/cloud_manager/refresh_parser.rb
+++ b/app/models/manageiq/providers/azure/cloud_manager/refresh_parser.rb
@@ -169,15 +169,32 @@ module ManageIQ::Providers
         process_collection(images, :vms) { |image| parse_managed_image(image) }
       end
 
+      # Collect marketplace image information if configured to do so. Normally
+      # users will specify images in their configuration file. If the option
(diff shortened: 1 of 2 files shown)


djberg96:
See also: https://github.com/ManageIQ/azure-armrest/issues/306

bronaghs:
@djberg96 - I think this warrants a spec test that covers the different variations of URNS

djberg96:
@bronaghs Understood, will write some specs soon.

djberg96:
@bronaghs Getting a spec failure for some reason, even though I can see in the parser that it's picking them up. Need a second pair of eyes when you get the chance please.

PR #125, opened 2017-09-13
Wait for SSA Snapshot Success

jerryk55:
We need to wait for success status from the SSA Snapshot
operations for both Managed and non-managed (blob) disks.
The Managed snapshot will use the Snapshot Service to wait
while the blob snapshot will use the Storage Account Service
to wait.

This is the story of the continuing saga of the BZs:
https://bugzilla.redhat.com/show_bug.cgi?id=1463780 (for a running non-managed disk VM)
and
https://bugzilla.redhat.com/show_bug.cgi?id=1475540 (for Managed Disk VMs)

This PR needs to be back ported to FINE and will be added to a hot fix covering both of these BZs.
Not needing to be back ported but still needed to be added are a PR to bump the manageiq-smartstate gem version to 1.4 and one in manageiq to use 1.4.

@roliveri please review for sanity (especially the length of the wait)
@bronaghs or @djberg96 please review and merge so we can put these blocker BZs to bed and push out the hot fix.  Thanks all.

### DIFF
--- a/app/models/manageiq/providers/azure/cloud_manager.rb
+++ b/app/models/manageiq/providers/azure/cloud_manager.rb
@@ -144,7 +144,7 @@ class ManageIQ::Providers::Azure::CloudManager < ManageIQ::Providers::CloudManag
     rescue ::Azure::Armrest::NotFoundException, ::Azure::Armrest::ResourceNotFoundException => err
       begin
         response = snap_svc.create(ssa_snap_name, resource_group, snap_options)
-        raise "Maximum snapshot wait time exceeded" unless snap_svc.wait(response.response_headers, SSA_SNAPSHOT_WAIT_TIME) == "Succeeded"
+        raise "Maximum snapshot wait time exceeded" unless snap_svc.wait(response.response_headers, SSA_SNAPSHOT_WAIT_TIME) =~ /^succe/i
         return ssa_snap_name
       rescue => err
         _log.error("vm=[#{vm.name}], error: #{err}")
@@ -160,7 +160,7 @@ class ManageIQ::Providers::Azure::CloudManager < ManageIQ::Providers::CloudManag
     _log.debug("vm=[#{vm.name}] creating SSA snapshot for #{vm.blob_uri}")
     begin
       snapshot_info = vm.storage_acct.create_blob_snapshot(vm.container, vm.blob, vm.key)
-      raise "Maximum snapshot wait time exceeded" unless vm.storage_acct_service.wait(snapshot_info, SSA_SNAPSHOT_WAIT_TIME) == "Succeeded"
+      raise "Maximum snapshot wait time exceeded" unless vm.storage_acct_service.wait(snapshot_info, SSA_SNAPSHOT_WAIT_TIME) =~ /^succe/i
       return snapshot_info[:x_ms_snapshot]
     rescue => err
       _log.error("vm=[#{vm.name}], error:#{err}")
--- a/app/models/manageiq/providers/azure/cloud_manager.rb
+++ b/app/models/manageiq/providers/azure/cloud_manager.rb
@@ -18,6 +18,7 @@ class ManageIQ::Providers::Azure::CloudManager < ManageIQ::Providers::CloudManag
 
   include ManageIQ::Providers::Azure::ManagerMixin
 
+  SSA_SNAPSHOT_WAIT_TIME = 1800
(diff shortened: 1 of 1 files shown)


app/models/manageiq/providers/azure/cloud_manager.rb
```diff
@@ -142,7 +143,8 @@ def vm_create_evm_managed_snapshot(vm)
       snap_svc.get(ssa_snap_name, resource_group)
     rescue ::Azure::Armrest::NotFoundException, ::Azure::Armrest::ResourceNotFoundException => err
       begin
-        snap_svc.create(ssa_snap_name, resource_group, snap_options)
+        response = snap_svc.create(ssa_snap_name, resource_group, snap_options)
+        raise "Maximum snapshot wait time exceeded" unless snap_svc.wait(response.response_headers, SSA_SNAPSHOT_WAIT_TIME) == "Succeeded"
```
djberg96: I'd change this to `=~ /^succe/i` because, depending on the operation, azure returns "Success" or "Succeeded", and I don't remember where and/or why.

app/models/manageiq/providers/azure/cloud_manager.rb
```diff
@@ -158,6 +160,7 @@ def vm_create_evm_blob_snapshot(vm)
     _log.debug("vm=[#{vm.name}] creating SSA snapshot for #{vm.blob_uri}")
     begin
       snapshot_info = vm.storage_acct.create_blob_snapshot(vm.container, vm.blob, vm.key)
+      raise "Maximum snapshot wait time exceeded" unless vm.storage_acct_service.wait(snapshot_info, SSA_SNAPSHOT_WAIT_TIME) == "Succeeded"
```
djberg96: Same, same.

jerryk55: ditto.

app/models/manageiq/providers/azure/cloud_manager.rb
jerryk55: @djberg96 cool - thanks for the heads up.

djberg96:
👍

djberg96:
@bronaghs Jerry is going to change the timeout handling.

jerryk55:
After offline discussion with @roliveri we changed the wait processing to allow the Job to control timeout rather than the wait.

jerryk55:
@miq-bot add_label fine/yes

bronaghs:
@roliveri - can you approve?

app/models/manageiq/providers/azure/cloud_manager.rb
```diff
@@ -142,7 +142,9 @@ def vm_create_evm_managed_snapshot(vm)
       snap_svc.get(ssa_snap_name, resource_group)
     rescue ::Azure::Armrest::NotFoundException, ::Azure::Armrest::ResourceNotFoundException => err
       begin
-        snap_svc.create(ssa_snap_name, resource_group, snap_options)
+        response = snap_svc.create(ssa_snap_name, resource_group, snap_options)
+        # wait a minute at a time, allowing the Job Timeout to handle long-running snapshots here
+        next until snap_svc.wait(response.response_headers) =~ /^succe/i
```
roliveri: This probably can be merged for now, so we can get the fix out. But this condition is too specific. If it doesn't complete successfully, we'll have to wait for the job to time out. The test should be more along the lines of:
```Ruby
next if snap_svc.wait(response.response_headers) =~ /^In progress/i
```
Not sure of the specific "in progress" state.

Same comment for the similar check below.

simaishi:
Fine backport details:

```
$ git log -1
commit 26ddc4fb61cfdaa1fcb697c47b23da2216ed3b02
Date:   Thu Sep 14 16:18:51 2017 -0400

    Merge pull request #125 from jerryk55/wait_for_snapshot_success
    
    Wait for SSA Snapshot Success
    (cherry picked from commit 9217377b32db9ad7251b30cba442ff8a0e0fb305)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1491310
```

app/models/manageiq/providers/azure/cloud_manager.rb
jerryk55: So I discussed that with @djberg96 on last Wednesday and he was of the opinion that there would be an Exception raised in the case of an unsuccessful completion and therefore we wouldn't hit the test regardless.

PR #128, opened 2017-09-20
Add logging of timings, counts, and memory usage

agrare:
Additional logging around collecting inventory specifically how long it took, how many items were returned, and the memory usage after each collection.

### DIFF
--- a/app/models/manageiq/providers/azure/cloud_manager/refresh_parser.rb
+++ b/app/models/manageiq/providers/azure/cloud_manager/refresh_parser.rb
@@ -55,25 +55,29 @@ module ManageIQ::Providers
       private
 
       def get_resource_groups
-        process_collection(resource_groups, :resource_groups) do |resource_group|
+        groups = collect_inventory(:resource_groups) { resource_groups }
+        process_collection(groups, :resource_groups) do |resource_group|
           parse_resource_group(resource_group)
         end
       end
 
       def get_series
-        series = []
-        begin
-          series = @vmm.series(@ems.provider_region)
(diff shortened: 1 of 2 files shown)


agrare:
Maybe the memory usage logging could go to debug instead of info?

bronaghs:
I think debug since once addressed we should not always need it, and it could clutter the logs.

djberg96:
@agrare Looks ok to me, though I agree it should probably be debug level info.

agrare:
Okay memory usage logged as debug now

app/models/manageiq/providers/azure/cloud_manager/refresh_parser.rb
```diff
@@ -125,7 +129,9 @@ def get_resource_status_message(resource)
 
       def get_stack_templates
         # download all template uris
+        _log.info("Retrieving templates...")
         @template_uris.each { |uri, template| template[:content] = download_template(uri) }
+        _log.info("Retrieving templates...Complete - Count [#{@template_uris.count}]")
```
bronaghs: @agrare - The downloading of orchestration templates is expensive, can we monitor the memory usage there too?

agrare: @bronaghs oh yeah definitely, I missed this one because it didn't use the usual inv collect method.

bronaghs:
@djberg96 - Can logging be added to the azure-armrest gem where we get private images? That code looks very expensive and I think we could do with some insight there.

PR #129, opened 2017-09-20
[FINE] Treat Resource Group as a String not an Object

jerryk55:
A previous change assumed that  PR https://github.com/ManageIQ/manageiq-providers-azure/pull/115 had been backported from Fine to upstream converting
Resource Groups to objects which had to be dereferenced to get to the name attribute.
This PR had not been backported so we need to treat the value as a String.

This is for high priority BZ https://bugzilla.redhat.com/show_bug.cgi?id=1475540
and goes hand in hand with https://github.com/ManageIQ/manageiq-gems-pending/pull/273

This is for FINE only.

@roliveri please review.  @djberg96 and/or @bronaghs please validate based on the above info and our discussions.


djberg96:
@jerryk55 I think the best thing to do is cover your bases:

    resource_group = vm.resource_group.try(:name) || vm.resource_group

jerryk55:
@djberg96 that is an option but if the .try clause works our tests will surely break as is...

PR #133, opened 2017-10-02
SSA Support for Managed Images

jerryk55:
Add support for SSA on Managed Images.  These
images may or may not have Managed Disks as well.

These changes are in support of blocker BZ https://bugzilla.redhat.com/show_bug.cgi?id=1496523.
Parallel changes have been added to the manageiq-smartstate gem by way of
PR https://github.com/ManageIQ/manageiq-smartstate/pull/31.

@roliveri please review for rationality with Smartstate.  @bronaghs and/or @djberg96 please review
as well and merge when appropriate.  Adding the Fine back port label as well.

### DIFF
--- a/app/models/manageiq/providers/azure/cloud_manager/vm_or_template_shared.rb
+++ b/app/models/manageiq/providers/azure/cloud_manager/vm_or_template_shared.rb
@@ -47,12 +47,11 @@ module ManageIQ::Providers::Azure::CloudManager::VmOrTemplateShared
   end
 
   def os_disk
-    if template 
-      @os_disk ||= image_object.properties.storage_profile.os_disk
-    else
-      @os_disk ||= vm_object.properties.storage_profile.os_disk
-    end
-    @os_disk
+    @os_disk ||= if template
+                   image_object.properties.storage_profile.os_disk
+                 else
+                   vm_object.properties.storage_profile.os_disk
+                 end
   end
 
   delegate :managed_disk?, to: :vm_object
@@ -63,7 +62,7 @@ module ManageIQ::Providers::Azure::CloudManager::VmOrTemplateShared
   end
 
   def managed_resource_group
-    return nil unless (ems_ref =~ /^\/subscriptions\//i)
+    return nil unless ems_ref =~ /^\/subscriptions\//i
     ref_parts = ems_ref.split('/')
     if ref_parts[3] =~ /resourceGroups/i
       return ref_parts[4]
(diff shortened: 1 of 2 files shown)


jerryk55:
@miq-bot add_label fine/yes

app/models/manageiq/providers/azure/cloud_manager/vm_or_template_shared.rb
```diff
@@ -37,17 +42,53 @@ def vm_object
...
+  def managed_image?
+    return false unless ems_ref =~ /^\/subscriptions\//i
+    os_disk.try(:managed_disk) ? true : false
+  end
+
+  def managed_resource_group
+    return nil unless ems_ref =~ /^\/subscriptions\//i
+    ref_parts = ems_ref.split('/')
+    if ref_parts[3] =~ /resourceGroups/i
+      return ref_parts[4]
```
djberg96: There should be a resource_group association on the Rails side for current code.

jerryk55: @djberg96 shoulda woulda coulda... :-)   If you recall I asked you about this because there wasn't a resource group.  Its nil.  And needs to work in Fine too, where there is no association.

djberg96: Hmmm...

app/models/manageiq/providers/azure/cloud_manager/vm_or_template_shared.rb
```diff
@@ -13,6 +13,11 @@ def storage_acct_service(connection = nil)
     @storage_acct_service ||= ::Azure::Armrest::StorageAccountService.new(@connection)
   end
 
+  def image_service(connection = nil)
+    @connection ||= connection || ext_management_system.connect
+    @image_service ||= ::Azure::Armrest::Storage::ImageService.new(@connection)
+  end
+
```
djberg96: We have something like this in the refresh_helper_methods.rb. Could that be mixed in here instead? That has the advantage of locking down the api-version, too.

jerryk55: @djberg96 while that is an interesting thought, I would be loathe to mix in all the other refresher-related methods from that file into this code.  Also since this PR is only adding the highlighted method, but we already have a couple of others as well, I'd probably rather add an issue tracking this, and pull all the "services" from both here and the refresh_helper_methods.rb into a separate mixin just defining the services.  Does that sound good?

chrispy1:
@miq-bot add_label blocker

app/models/manageiq/providers/azure/cloud_manager/vm_or_template_shared.rb
djberg96: @jerryk55 Ok, let's refactor that in the future.

simaishi:
Fine backport details:

```
$ git log -1
commit 1394b34298d177e2aaa585ec591e7614cc539db7
Date:   Tue Oct 3 15:05:15 2017 -0600

    Merge pull request #133 from jerryk55/ssa_azure_managed_images
    
    SSA Support for Managed Images
    (cherry picked from commit 9b0c67a6774aefd79b416fd71cbfe8bdff324794)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1498556
```

PR #134, opened 2017-10-03
Update azure-armrest to 0.8.4

djberg96:
This PR updates the gemspec by updating the azure-armest gem from 0.8.3 to 0.8.4. The 0.8.4 release notably reduces memory in the `StorageAccountService#list_all_private_images` method. There's also a logging fix.


simaishi:
Backported to Fine via https://github.com/ManageIQ/manageiq-providers-azure/pull/138

PR #135, opened 2017-10-03
fix factory orchestration_stack for azure

bzwei:
Point the template to use the new factory.
The old one has been removed from manageiq repo.


bzwei:
@miq-bot add_label bug
@miq-bot assign @djberg96

PR #142, opened 2017-10-20
Always collect all resource groups

djberg96:
I accidentally introduced an issue in https://github.com/ManageIQ/manageiq-providers-azure/pull/72 when we created an actual resource_group association on the Rails side instead of relying on a string property of the object.

The fundamental problem is that Azure does not require resources to be in the same location as their resource group. Consequently, if you try to access `vm.resource_group` it could come back nil because the resource group was never collected by the refresh parser because it's in a different region.

Because Azure does not enforce this, and because our own resource_group table doesn't store location information, this PR alters the refresh parser so that it collects ALL resource groups regardless of provider location.

I've added some specs as well, creating a VM with a location that is different than the resource group that it resides in.

BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1503917


djberg96:
@miq-bot add_label fine/no

PR #143, opened 2017-10-20
Set the location property of managed disks to ensure disk saving

bronaghs:
This fixes a bug where managed disks are saved to the database when a provider is first added but are deleted upon subsequent refreshes of the same provider.

### Background
The saving of instance disks involves creating an array of disks already associated with the VM from the previous refresh, and one by one, each disk is removed as it is saved. What is left is old disks that no longer exist on the VM and will be deleted.
This array of pre-existing disks is created from the VM.disks association which is then indexed using the TypedIndex class.
This indexing uses the `location` property of a disk as a key. If the location is missing, as is the case with managed disks,  that disk will no longer be associated with the instance, will remain in the array of pre-existing disks and will eventually be deleted, see [here](https://github.com/ManageIQ/manageiq/blob/master/app/models/ems_refresh/save_inventory.rb#L261..#L272) and  [here](https://github.com/ManageIQ/manageiq/blob/master/app/models/ems_refresh/save_inventory_helper.rb#L94...#L105).
## Steps to reproduce

- Add an Azure provider - Make sure at lease one VM has managed data disks.
- Note the managed disks are saved to the DB
- Refresh the same provider
- Note the managed data disk was deleted.

After this PR is merged, the managed data disks will persist after a second refresh.

https://bugzilla.redhat.com/show_bug.cgi?id=1503866


app/models/manageiq/providers/azure/cloud_manager/refresh_parser.rb
```diff
@@ -308,6 +308,7 @@ def populate_hardware_hash_with_disks(hardware_disks_array, instance)
           disk_size      = disk.respond_to?(:disk_size_gb) ? disk.disk_size_gb * 1.gigabyte : 0
           disk_name      = disk.name
           disk_location  = disk.try(:vhd).try(:uri)
+          disk_location  = disk.try(:vhd).try(:uri) || disk.try(:managed_disk).try(:id)
```
djberg96: Did you mean to drop the previous line?

bronaghs:
@djberg96 - yes.

djberg96:
Just thinking out loud here, we run the refresh `2.times` right now within the specs, so I'm surprised the current specs didn't pick it up. Were we just not looking for that in particular?

bronaghs:
@djberg96 - The reason why the spec passes is because it is testing the existence of an OS disk, not a data disk:
https://github.com/ManageIQ/manageiq-providers-azure/blob/master/spec/models/manageiq/providers/azure/cloud_manager/refresher_spec.rb#L14

djberg96:
@bronaghs oh, good catch!

chessbyte:
Looks good to me!  👍

app/models/manageiq/providers/azure/cloud_manager/refresh_parser.rb
```diff
@@ -307,7 +307,7 @@ def populate_hardware_hash_with_disks(hardware_disks_array, instance)
         data_disks.each do |disk|
           disk_size      = disk.respond_to?(:disk_size_gb) ? disk.disk_size_gb * 1.gigabyte : 0
           disk_name      = disk.name
-          disk_location  = disk.try(:vhd).try(:uri)
+          disk_location  = disk.try(:vhd).try(:uri) || disk.try(:managed_disk).try(:id)
```
durandom: `||` will get naughty if the first method exists but returns an empty string... 
We had that problem 3 times in amazon.

But I guess the azure sdk does a better job here :)

simaishi:
@bronaghs Please create a Fine PR for this. Spec files are conflicting a lot, so it's better that gets resolved by you.

PR #148, opened 2017-10-30
Pass resource group as a string to the instance delete method

bronaghs:
The instance delete operation passes a resource group object to the armrest gem delete function but should pass a string representation.
This was introduced [here](https://github.com/ManageIQ/manageiq/pull/14000) 

https://bugzilla.redhat.com/show_bug.cgi?id=1507574


djberg96:
LGTM

PR #158, opened 2017-11-03
Correct and update disk information

djberg96:
We inadvertently used the host disk (flavor) size when we really wanted to use the OS/data disk size. This updates it so that it only falls back to flavor size if it can't find the OS/data disk size.

In addition, we need to include the `disk_type` (managed or unmanaged) and storage type (standard LRS, premium, etc) for the `mode`.


bronaghs:
I'm not sure we should even fall back to the flavor disk size if that is the _maximum_ size that can be allocated to the OS Disk, not the _actual_.

djberg96:
@bronaghs ok, will just default to 0 if it doesn't have the expected property then.

blomquisg:
Looks like this need VCR updates

app/models/manageiq/providers/azure/cloud_manager/refresh_parser.rb
```diff
@@ -350,13 +357,54 @@ def populate_hardware_hash_with_series_attributes(hardware_hash, instance, serie
...
-        os_disk = instance.properties.storage_profile.os_disk
-        sz      = series[:root_disk_size]
-        vhd_loc = os_disk.try(:vhd).try(:uri) || os_disk.try(:managed_disk).try(:id)
+        add_azure_instance_disk(hardware_hash[:disks], instance)
+
+        # No data available on swap disk? Called temp or resource disk.
+      end
+
+      def add_azure_instance_disk(array, instance, os_disk = nil)
+        os_disk ||= instance.properties.storage_profile.os_disk
```
bronaghs: the right side will only parse for OS Disks, can you pass in the OS Disk like you pass in the data disk so this isn't needed?

app/models/manageiq/providers/azure/cloud_manager/refresh_parser.rb
```diff
@@ -350,13 +357,54 @@ def populate_hardware_hash_with_series_attributes(hardware_hash, instance, serie
...
+
+            storage_acct = @storage_accounts.find { |s| s.name.casecmp(storage_name).zero? }
+            mode = storage_acct.sku.name
+
+            if disk_size.nil?
+              storage_keys = @sas.list_account_keys(storage_acct.name, storage_acct.resource_group)
+              storage_key  = storage_keys['key1'] || storage_keys['key2']
+              blob_props   = storage_acct.blob_properties(container_name, blob_name, storage_key)
+              disk_size    = blob_props.content_length.to_i
+            end
```
bronaghs: why would disk size be nil? I would love to avoid the complexities around lines 389 - 392

app/models/manageiq/providers/azure/cloud_manager/refresh_parser.rb
```diff
@@ -350,13 +357,54 @@ def populate_hardware_hash_with_series_attributes(hardware_hash, instance, serie
...
+          :device_name     => os_disk.name,
+          :location        => disk_location,
+          :size            => disk_size,
+          :disk_type       => disk_type,
+          :mode            => mode
+        }
 
-        # No data availbale on swap disk? Called temp or resource disk.
+        array << disk
       end
```
bronaghs: @djberg96 
Until I dug deep, it looked like this method `add_azure_instance_disk` accepts a data disk as the third parameter yet only parses OS disks. Upon closer review, it looks like data disks will be parsed correctly. Can you rename the `osDisk` variable to `disk` to avoid confusion?

app/models/manageiq/providers/azure/cloud_manager/refresh_parser.rb
```diff
@@ -350,13 +357,54 @@ def populate_hardware_hash_with_series_attributes(hardware_hash, instance, serie
...
-        os_disk = instance.properties.storage_profile.os_disk
-        sz      = series[:root_disk_size]
-        vhd_loc = os_disk.try(:vhd).try(:uri) || os_disk.try(:managed_disk).try(:id)
+        add_azure_instance_disk(hardware_hash[:disks], instance)
+
+        # No data available on swap disk? Called temp or resource disk.
+      end
+
+      def add_azure_instance_disk(array, instance, os_disk = nil)
+        os_disk ||= instance.properties.storage_profile.os_disk
```
bronaghs: Why did you insert the string `azure` into the method name? Its implied that this applies to Azure.

app/models/manageiq/providers/azure/refresh_helper_methods.rb
```diff
@@ -64,6 +64,11 @@ def resource_groups
     @resource_groups ||= @rgs.list
   end
 
+  def resources_for(resource_type)
+    filter = "resourceType eq '#{resource_type}'"
+    @res.list_all(:filter => filter)
+  end
+
```
bronaghs: is this method used?

app/models/manageiq/providers/azure/refresh_helper_methods.rb
```diff
@@ -132,6 +137,12 @@ def network_security_group_service(config)
     end
   end
 
+  def resource_service(config)
+    ::Azure::Armrest::ResourceService.new(config).tap do |service|
+      service.api_version = Settings.ems.ems_azure.api_versions.resource.to_s
+    end
+  end
```
bronaghs: is this method used?

app/models/manageiq/providers/azure/cloud_manager/refresh_parser.rb
djberg96: I haven't completely nailed down why some vm's include the disk size in their storage profile, and why some don't, but it must have something to do with the way they were provisioned, or possibly even *when* they were provisioned. Or it could have something to do with the OS. In any case, I cannot predict it at this time.

You can see the difference by running the following commands on the CLI:
```
az vm show -n miq-test-win12 -g miq-azure-test1 # This doesn't show a disk size.
az vm show -n dberger1 -g dberger1 # This does show a disk size
```

app/models/manageiq/providers/azure/refresh_helper_methods.rb
djberg96: Oh, I thought I removed it. The original plan was to use the ResourceService, until I discovered that it did not include the information that I needed.

app/models/manageiq/providers/azure/refresh_helper_methods.rb
djberg96: Ditto as above, will remove.

app/models/manageiq/providers/azure/cloud_manager/refresh_parser.rb
djberg96: I can rename it. I didn't want to use `add_instance_disk` since it's inherited, but I guess there's no harm in just redefining it.

app/models/manageiq/providers/azure/cloud_manager/refresh_parser.rb
djberg96: Sure thing.

app/models/manageiq/providers/azure/cloud_manager/refresh_parser.rb
bronaghs: Yea you could override it.

djberg96:
@bronaghs Ok, I think I've made all the requested changes.

app/models/manageiq/providers/azure/cloud_manager/refresh_parser.rb
```diff
@@ -350,13 +353,55 @@ def populate_hardware_hash_with_series_attributes(hardware_hash, instance, serie
...
-        vhd_loc = os_disk.try(:vhd).try(:uri) || os_disk.try(:managed_disk).try(:id)
+        add_instance_disk(hardware_hash[:disks], instance)
+
+        # No data available on swap disk? Called temp or resource disk.
+      end
+
+      # Redefine the inherited method for our purposes
+      def add_instance_disk(array, instance, disk = nil)
+        disk ||= instance.properties.storage_profile.os_disk
+
```
bronaghs: can you pass the OS Disk in as a parameter so its the same as you do for data disks? This line is OS disk specific yet the method is used for all disks.

app/models/manageiq/providers/azure/cloud_manager/refresh_parser.rb
```diff
@@ -350,13 +353,55 @@ def populate_hardware_hash_with_series_attributes(hardware_hash, instance, serie
         hardware_hash[:memory_mb]       = series[:memory] / 1.megabyte
         hardware_hash[:disk_capacity]   = series[:root_disk_size] + series[:swap_disk_size]
 
-        os_disk = instance.properties.storage_profile.os_disk
-        sz      = series[:root_disk_size]
-        vhd_loc = os_disk.try(:vhd).try(:uri) || os_disk.try(:managed_disk).try(:id)
+        add_instance_disk(hardware_hash[:disks], instance)
+
+        # No data available on swap disk? Called temp or resource disk.
```
bronaghs: Can you delete this comment?

app/models/manageiq/providers/azure/cloud_manager/refresh_parser.rb
```diff
@@ -350,13 +353,55 @@ def populate_hardware_hash_with_series_attributes(hardware_hash, instance, serie
         hardware_hash[:memory_mb]       = series[:memory] / 1.megabyte
         hardware_hash[:disk_capacity]   = series[:root_disk_size] + series[:swap_disk_size]
```
bronaghs: Is the disk capacity correct? I do not think we should get this value from the series.

bronaghs:
I think, in a separate PR, a disk should be added for a resource disk. Thoughts?
A description of a resource disk:

> Resource disk is a locally attached disk, typically an SSD 
> It is a host disk (or rather part of one on the host depending on the VM size) and is not backed by any storage account. The contents of this drive is lost upon host reallocation. Agent creates a temp drive on it with a warning txt file with this message.
>

app/models/manageiq/providers/azure/cloud_manager/refresh_parser.rb
djberg96: I would have to know the difference between size and disk_capacity. And that's not part of the disks table. It may actually be correct.

app/models/manageiq/providers/azure/cloud_manager/refresh_parser.rb
djberg96: Done.

app/models/manageiq/providers/azure/cloud_manager/refresh_parser.rb
djberg96: Ok, done.

bronaghs:
@djberg96 Spoke to PM (Brad Ascar). He recommends adding a setting to the settings.yml file that allows the user to enable or disable setting the tier on unmanaged disks. Similar to how we enable/disable the gathering of public images. The default would be true. A documentation BZ would have to be filed describing this.

bronaghs:
👍 
Squash and then I'll merge once green

simaishi:
@djberg96 Tried to backport to Fine branch, but  `spec/vcr_cassettes/manageiq/providers/azure/cloud_manager/refresher.yml` is conflicting. Should I take the yml file from this PR as is, ignoring all the conflicts?

djberg96:
@simaishi yes, just use the .yml file from this PR please.

simaishi:
Fine backport details:

```
$ git log -1
commit b19e9c1844c16ece5c8bcb352ff15eb06f545e31
Date:   Thu Nov 9 16:40:21 2017 -0500

    Merge pull request #158 from djberg96/disk_information
    
    Correct and update disk information
    (cherry picked from commit f8d65fcd1ca49ab286e980d62f2a9ad2d62ae7b2)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1512728
```

simaishi:

```
$ git log -1
commit e29a5fb43f8c3f55baca86cf8f2d60660a630e26
Date:   Thu Nov 9 16:40:21 2017 -0500

    Merge pull request #158 from djberg96/disk_information
    
    Correct and update disk information
    (cherry picked from commit f8d65fcd1ca49ab286e980d62f2a9ad2d62ae7b2)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1512727
```

PR #161, opened 2017-11-06
Fix exception handing for credential validation on raw_connect

tumido:
When provider is added the credentials validation is newly done via `raw_connect` (class method) instead of `verify_credentials` (needed EMS object). Further reading on this change can be found here: ManageIQ/manageiq-ui-classic#1580

This change ensures that exceptions are handled properly, because the `verify_credentials` is not called anymore.

### DIFF
--- a/app/models/manageiq/providers/azure/manager_mixin.rb
+++ b/app/models/manageiq/providers/azure/manager_mixin.rb
@@ -10,19 +10,9 @@ module ManageIQ::Providers::Azure::ManagerMixin
   end
 
   def verify_credentials(_auth_type = nil, options = {})
-    require 'azure-armrest'
-    conf = connect(options)
-  rescue ArgumentError => err
-    raise MiqException::MiqInvalidCredentialsError, _("Incorrect credentials - #{err.message}")
-  rescue ::Azure::Armrest::UnauthorizedException, ::Azure::Armrest::BadRequestException
-    raise MiqException::MiqInvalidCredentialsError, _("Incorrect credentials - check your Azure Client ID and Client Key")
-  rescue MiqException::MiqInvalidCredentialsError
-    raise # Raise before falling into catch-all block below
-  rescue => err
-    _log.error("Error Class=#{err.class.name}, Message=#{err.message}, Backtrace=#{err.backtrace}")
-    raise MiqException::MiqInvalidCredentialsError, _("Unexpected response returned from system: #{err.message}")
-  else
-    conf
+    connection_rescue_block do
+      connect(options)
+    end
   end
 
   module ClassMethods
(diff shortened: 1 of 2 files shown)


tumido:
cc @jntullo 
@miq-bot add_label bug gaprindashvili/yes

jntullo:
Thanks for the fix @tumido :1st_place_medal:

app/models/manageiq/providers/azure/manager_mixin.rb
```diff
@@ -10,19 +10,9 @@ def connect(options = {})
...
-  rescue ::Azure::Armrest::UnauthorizedException, ::Azure::Armrest::BadRequestException
-    raise MiqException::MiqInvalidCredentialsError, _("Incorrect credentials - check your Azure Client ID and Client Key")
-  rescue MiqException::MiqInvalidCredentialsError
-    raise # Raise before falling into catch-all block below
-  rescue => err
-    _log.error("Error Class=#{err.class.name}, Message=#{err.message}, Backtrace=#{err.backtrace}")
-    raise MiqException::MiqInvalidCredentialsError, _("Unexpected response returned from system: #{err.message}")
-  else
-    conf
+    connection_rescue_block do
```
blomquisg: Do you need to call `connection_rescue_block` here?

It looks like `connect` will call `raw_connect`, which already wraps the call to `Configuration.new` with `connection_rescue_block`.

tumido: that's correct, thanks!

juliancheal:
LGTM 👍

djberg96:
This appears to break the specs, which return the error we expect. With this PR as written, it doesn't return the error we expect.

juliancheal:
@djberg96 I couldn't get the specs to run locally, so wasn't sure if/what would break.

tumido:
@djberg96 what about this?

djberg96:
@tumido it seems that now you're really just testing the azure-arrmest gem behavior rather than our cloud manager.

tumido:
@djberg96 hm, I don't think so. The call stack in the test is:
1. `verify_credentials`
2. `connect`
3. `raw_connect`
4. `connection_rescue_block`
5. `Azure::Armrest::Configuration.new`

So if we want to test `verify_credentials` to raise exception and the exception handler is in `connection_rescue_block`, I would do that by crippling the `Azure::Armrest::Configuration.new`. I'm not getting anything from azure-arrmest here.

blomquisg:
Yeah, I don't think this is testing the armrest behavior.  I think it's mocking the armrest behavior, and testing what happens in the CloudManager depending on the mocked armrest behavior.

djberg96:
@blomquisg Ah, I see now. Ok.

bronaghs:
Do we still use the [verify_credentials](https://github.com/ManageIQ/manageiq-providers-azure/blob/master/app/models/manageiq/providers/azure/manager_mixin.rb#L12) method?

@jntullo @h-kataria - In the [UI code](https://github.com/ManageIQ/manageiq-ui-classic/blob/master/app/controllers/mixins/manager_controller_mixin.rb#L78..#L93), is this used?

Would this method hang around for the sake of the REST API?

simaishi:

```
$ git log -1
commit bb2d567bce12f4c865af653f9a8c3a69871d5681
Date:   Mon Nov 6 12:06:01 2017 -0700

    Merge pull request #161 from tumido/fix_cred_validation
    
    Fix exception handing for credential validation on raw_connect
    (cherry picked from commit 09f5782497d812f1c5f06f815c36ad6ec02a86e7)
```

PR #168, opened 2017-11-20
Don't include manageiq-smartstate gem in gemspec here

jerryk55:
The manageiq-smartstate gem version 0.2.3 was previously included in this gemspec
but since the gem is also included in the main ManageIQ repo Gemfile at the greatest
0.2 version the line in this provider repo is extraneous and incorrectly ties us to
a lower version than necessary.  Version 0.2.4 was just released and removing the
gem version from this gemspec file allows us to use the newest version.

This is in support of PR https://github.com/ManageIQ/manageiq-smartstate/pull/45
which contains a fix for BZ https://bugzilla.redhat.com/show_bug.cgi?id=1490488.

@roliveri @hsong-rh and maybe @djberg96 please review and merge when appropriate.  Thanks.

### DIFF
--- a/manageiq-providers-azure.gemspec
+++ b/manageiq-providers-azure.gemspec
@@ -14,7 +14,6 @@ Gem::Specification.new do |s|
   s.files = Dir["{app,config,lib}/**/*"]
 
   s.add_dependency "azure-armrest", "~>0.9.3"
-  s.add_dependency "manageiq-smartstate", "~>0.2.3"
 
   s.add_development_dependency "codeclimate-test-reporter", "~> 1.0.0"
   s.add_development_dependency "simplecov"


jerryk55:
@miq-bot add_label fine/yes

jerryk55:

jerryk55:
@miq-bot add_label bug

jerryk55:
@miq-bot add_label smartstate

djberg96:
👍

simaishi:
@jerryk55 So we won't be updating Gemfile in `manageiq` repo to 0.2.3/0.2.4, to ensure the minimum version required to get the fix?

@Fryguy I often see people updating the minimum version in Gemfile/gemspec for cases like this. Do we have a policy around this?

jerryk55:
@simaishi the version in the Gemfile picks up the latest 0.2 version so when I did a bundle update it switched from 0.2.3 to 0.2.4.

simaishi:
@jerryk55 Yes, I know you'll get the latest if you bundle update. But I thought setting the minimum is to ensure everyone will get the fix. If you have 0.2.1, nothing will complain as it's > 0.2, but you actually won't have a fix for this issue.

Also note, this approach won't work for downstream z-stream releases where we lock gem versions. If the current gem version meets the requirement, it won't be updated.

jerryk55:
@simaishi ok if you need me to make that Gemfile change I have no problem - it sounds like that's the case, right?

simaishi:
We're not using this gem in `fine` branch, so this particular change will go to Fine branch without problem (backport manageiq-smartstate PR to gems-pending).

But it would be good to know what our policy is, so we know what to expect in the future...

simaishi:

```
$ git log -1
commit 652ef41f9950e75ec06d731236c7089d84ff8db7
Date:   Mon Nov 20 14:03:06 2017 -0700

    Merge pull request #168 from jerryk55/allow_manageiq-smartstate_0.2.4_and_beyond
    
    Don't include manageiq-smartstate gem in gemspec here
    (cherry picked from commit c41104db23b9d5174b5a5b5d51852d2ab22e20c1)
```

simaishi:
Removed `fine/yes` as the PR that added the gem in gemspec was never backported to Fine branch.

Fryguy:
If this code depends on smartstate directly, it should be in the gemspec.  Typically, we keep loose restrictions in plugins, so this should be `s.add_dependency "manageiq-smartstate", "~>0.2"`

PR #170, opened 2017-11-22
Add i18n string placeholders test

mzazrivec:
This is just invocation of standard i18n placeholder test, included from main ManageIQ
repository. It's the same thing that we do in ui-classic to test that the provided translations
honor placeholders in the original (English) strings.


djberg96:
Should there be some actual specs here?

mzazrivec:
The spec in question already lives in core manageiq repo. This PR just includes that shared
example so that the test is run for the gettext catalogs in this repo.

djberg96:
@mzazrivec Alright, thanks. LGTM.

PR #171, opened 2017-11-22
Select only the event fields that we need

djberg96:
The event record returned by Azure contains lots of information that we aren't using. An individual json event record is about 2.5k. By selecting only the fields we need we can reduce memory usage down to less than 1k per record.

I've also added some specs for the EventParser.


bronaghs:
Nice @djberg96 !

djberg96:
@bronaghs Thanks. I accidentally discovered it while resolving your TODO. :)

djberg96:
@bronaghs Did I break something? Only just now noticed this:

https://gist.github.com/djberg96/ad3ab6c4e4af09fb6bdfa8e5e04a6980

djberg96:
@bronaghs Yep, here's the fix: https://github.com/ManageIQ/manageiq-providers-azure/pull/182

PR #172, opened 2017-11-22
[WIP] Allow for possibility of empty NIC

djberg96:
Handle the possibility that the user may not want to assign an IP to a VM when provisioning.

WIP for now until I know for sure what the UI option is.

https://bugzilla.redhat.com/show_bug.cgi?id=1497202


app/models/manageiq/providers/azure/cloud_manager/provision/cloning.rb
```diff
@@ -82,13 +80,22 @@ def prepare_for_clone_task
...
-          :networkInterfaces => [{:id => nic_id}],
         }
       }
     }
 
+    # The -1 value is set in ProvisionWorkflow to distinguish between the
+    # desire for a new Public IP address vs a private IP.
+    #
+    if floating_ip
+      nic_id = associated_nic || create_nic(true)
```
bronaghs: @djberg96 - In what scenario will `create_nic` on this line be called? Looks like this line will only be executed if floating_ip is not nil, which means a valid floating IP was selected and there is no need to create a new one.

app/models/manageiq/providers/azure/cloud_manager/provision/cloning.rb
```diff
@@ -82,13 +80,22 @@ def prepare_for_clone_task
...
+    else
+      public_ip = options[:floating_ip_address].first == -1
+      nic_id = create_nic(public_ip)
```
bronaghs: so the `else` statement will be executed if floating_ip is `nil` which would be either:
- the floating_ip dropdown was ignored altogether 
or
- "New" was selected and therefore the look up in the options helper returns `nil` - refer to lookup [here](https://github.com/ManageIQ/manageiq/blob/master/app/models/manageiq/providers/cloud_manager/provision/options_helper.rb#L19)

app/models/manageiq/providers/azure/cloud_manager/provision/cloning.rb
```diff
@@ -82,13 +80,22 @@ def prepare_for_clone_task
...
       }
     }
 
+    # The -1 value is set in ProvisionWorkflow to distinguish between the
+    # desire for a new Public IP address vs a private IP.
+    #
+    if floating_ip
+      nic_id = associated_nic || create_nic(true)
+    else
+      public_ip = options[:floating_ip_address].first == -1
```
bronaghs: @djberg96 - Can you point me to where the options hash gets populated? What does this line mean?

app/models/manageiq/providers/azure/cloud_manager/provision/cloning.rb
```diff
@@ -82,13 +80,22 @@ def prepare_for_clone_task
...
           }
-        },
-        :networkProfile  => {
-          :networkInterfaces => [{:id => nic_id}],
         }
       }
     }
 
+    # The -1 value is set in ProvisionWorkflow to distinguish between the
+    # desire for a new Public IP address vs a private IP.
```
bronaghs: Is the word "private" correct here?

djberg96: Not sure what you're asking. If a Public IP isn't assigned, Azure will automatically assign a private IP.

app/models/manageiq/providers/azure/cloud_manager/provision/cloning.rb
djberg96: This was something the specs caught. It's possible that a public IP exists, but isn't actually associated with a NIC yet.

app/models/manageiq/providers/azure/cloud_manager/provision/cloning.rb
bronaghs: Ok, thanks

app/models/manageiq/providers/azure/cloud_manager/provision/cloning.rb
djberg96: It's coming from the UI. The line in question is https://github.com/djberg96/manageiq-providers-azure/blob/70fbc2f62b293ad1490a1530b98ae28473e6632a/app/models/manageiq/providers/azure/cloud_manager/provision_workflow.rb#L30

app/models/manageiq/providers/azure/cloud_manager/provision/cloning.rb
djberg96: Correct.

app/models/manageiq/providers/azure/cloud_manager/provision/cloning.rb
djberg96: In reality, just poking around on the Azure portal, it appears that NIC's with a public IP still have a private IP as well.

app/models/manageiq/providers/azure/cloud_manager/provision/cloning.rb
bronaghs: Ah I see. So this means when the user selects a floating ip from the drop down that is not associated with a NIC  we create a new IP called "<dest_name>-PublicIp" which isn't what the user intended. 
This is what we have always done, I wonder should we at least add some info logging since we are doing something the user didn't intend?

app/models/manageiq/providers/azure/cloud_manager/provision/cloning.rb
bronaghs: I was looking for the code that populates the options hash with all the valid floating Ips.
Does this line return what the user selecting from the floating_ip drop down?

app/models/manageiq/providers/azure/cloud_manager/provision/cloning.rb
djberg96: The other option is to just create the NIC dynamically if it's not associated with a NIC, and then associate it with the newly created NIC. Excluding them would be easier, though.

In any case, I think that's something that should be in a separate PR. I don't get the impression that people create public IP's independently of a NIC much in practice.

app/models/manageiq/providers/azure/cloud_manager/provision/cloning.rb
djberg96: I'm not sure where it's populated originally, somewhere on the UI side. Yes, that line returns what the user selected as a key/value pair on a typical Rails dropdown menu. I interjected the -1 there for us to distinguish between "create new" and "not selected".

I would have preferred to use a string key to be more descriptive, but the UI code blew up when it tried to sort it.

bronaghs:
Can the key to the "New" value also be "New" instead of -1?

djberg96:
Nope, the UI  blew up when I tried that due to some sorting code. I started with that. :)

bronaghs:
Yea the key is the floating IP Id which is an integer.

app/models/manageiq/providers/azure/cloud_manager/provision/cloning.rb
bronaghs: yeah it has always been there so i think its fine if it is addressed in a separate PR.

app/models/manageiq/providers/azure/cloud_manager/provision/cloning.rb
bronaghs: So the reason I was looking for the code that populates the options hash is because this line returns an array of floating ips, so its a multi-select? Will `first` always be -1?

djberg96: Oops I misspoke. The `options[:floating_ip_address]` will be a two element array of the currently selected item, i.e. [-1, "New"]. If the user actually selects something, then it will be something else.

bronaghs:
Looks great @djberg96 !

simaishi:

```
$ git log -1
commit e58c0038317b883ff512f2ad25133c02d5fce68a
Date:   Tue Dec 19 12:13:59 2017 -0500

    Merge pull request #172 from djberg96/optional_nic
    
    Allow for possibility of private IP
    (cherry picked from commit 9b9fba302552a6d0cbceeb98b4620d02c8b3f3ee)
    
    Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1530728
```

simaishi:
Fine backport details:

```
$ git log -1
commit e63af8e39e0d9283bdf42d5b3c67bad5ba379be7
Date:   Tue Dec 19 12:13:59 2017 -0500

    Merge pull request #172 from djberg96/optional_nic
    
    Allow for possibility of private IP
    (cherry picked from commit 9b9fba302552a6d0cbceeb98b4620d02c8b3f3ee)
    
    Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1531256
```

PR #174, opened 2017-11-28
Use include instead of have_attributes to avoid warnings

djberg96:
At the moment we're seeing a couple warnings when running the test suite, that are coming from our custom have_attributes definition. 

`WARNING: Use of 'have_attributes' with array access (:[]) is deprecated and will be removed shortly.
If you're matching attributes in hashes, use appropriate hash matchers instead ('include', 'eq').`

This PR just switches `have_attributes` to `include` for a couple specs.


bronaghs:
@djberg96 - Were these warnings always there? Or did something change with rspec?

djberg96:
@bronaghs It's coming from a custom `have_attributes` definition that we implemented at some point:

https://github.com/ManageIQ/manageiq/blob/c0c43f03a56d0933383cb52ad763ae71089adfa5/spec/support/custom_matchers/have_attributes.rb

PR #179, opened 2017-12-01
Update VCR cassettes

djberg96:
A recent api-version update from Azure broke our specs. This updates the cassettes with the current api-version string.

PR #183, opened 2017-12-07
Set api-version explicitly for discovery and regenerate cassettes

djberg96:
This is a followup to https://github.com/ManageIQ/manageiq-providers-azure/pull/179. It turns out the Service class we were using for discovery code did not explicitly set the api-version. This caused the requests to use the latest api-version, instead of the one we want locked in. This corrects that, and regenerates the VCR cassettes.

A separate PR will have to be generated for Fine/Euwe, because it uses a different service class.


simaishi:

```
$ git log -1
commit 577cd610f34604b1b4c29327f1d032720be75c09
Date:   Fri Dec 8 09:46:21 2017 -0500

    Merge pull request #183 from djberg96/spec_api_version
    
    Set api-version explicitly for discovery and regenerate cassettes
    (cherry picked from commit b3d018eebeb9307350f0390c17ef5b231a783ee9)
```

PR #180, opened 2017-12-04
Handle possibility of arrays for Network Security Groups

djberg96:
This handles an api-version update for Network Security Groups that started with api-version 2017-10-01. As of that date, multiple ports and/or ranges are returned as an array with a slightly different property name, e.g. `source_address_prefixes` instead of `source_address_prefix`.

For earlier versions of the api-version, it's possible neither property will exist (it doesn't handle arrays), in which case we simply return nil.

Note that for `port` and `end_port` I simply took the low and high values, respectively, since the underlying `firewall_rules` table isn't designed to handle arrays of values.

https://bugzilla.redhat.com/show_bug.cgi?id=1520196

PR #186, opened 2017-12-11
Update api-version string settings

djberg96:
This updates the api-version strings to the latest version for each service. We typically want to have each release on the latest api-version string, not only because new information may have been added or changed, but also because Azure eventually drops support for older api-versions.

A breakdown of the individual changes are as follows. Note that I am only looking at the output for individual records.

* AvailabilitySetService: No noticeable difference.
* NetworkSecurityGroupService: Now supports arrays of port ranges, prefixes, etc, in certain regions.
* IpAddressService: Sku information added.
* LoadBalancerService: Sku information added.
* Storage::ImageService: No noticeable difference.
* Network::NetworkInterfaceService: Adds a `virtualNetworkTapProvisioningState` property.
* ResourceService: No noticeable difference.
* ResourceGroupService: No noticeable difference.
* StorageAccountService: Adds the `networkAcls` property.
* TemplateDeploymentService: The ID elements of `outputResources` now contains a full ID instead of an abbreviated ID.
* VirtualMachineService: No noticeable difference.
* VirtualNetworkService: Adds the `enableDdosProtection` and `enableVmProtection` boolean properties.

Most of these changes will have no bearing on our current code. The change to NetworkSecurityGroupService was already handled by https://github.com/ManageIQ/manageiq-providers-azure/pull/180.

The only change that *might* be of concern is the TemplateDeploymentService change, but I don't see `outputResources` being referenced directly anywhere.


djberg96:
@bzwei Can you confirm that the change to `outputResources` in the `TemplateDeploymentService` won't affect us?

bronaghs:
Great. I will wait for @bzwei 's approval on the template piece before merging

djberg96:
@bronaghs I took the thumbs-up from @bzwei to mean that we were clear. :)

bronaghs:
Ah, you would be correct. 
Specs are failing, from a brief look, it seems that you need to regenerate cassettes.

PR #195, opened 2018-01-09
Only create new Public IP if one cannot be found

djberg96:
At the moment, if you select a public IP address when provisioning, but that IP is not currently associated with a NIC, it will create a new IP instead of using the one selected.

This PR modifies the provisioning code so that when a new NIC is provisioned, it will attempt to use the selected public IP before resorting to creating a new one.

Addresses https://bugzilla.redhat.com/show_bug.cgi?id=1531099

### DIFF
--- a/app/models/manageiq/providers/azure/cloud_manager/provision/cloning.rb
+++ b/app/models/manageiq/providers/azure/cloud_manager/provision/cloning.rb
@@ -139,7 +139,18 @@ module ManageIQ::Providers::Azure::CloudManager::Provision::Cloning
 
       if with_public_ip
         ips = ::Azure::Armrest::Network::IpAddressService.new(azure)
-        ip  = ips.create("#{dest_name}-publicIp", resource_group.name, :location => region)
+
+        # Use the existing Public IP if possible. Otherwise create a new one.
+        if floating_ip.try(:ems_ref)
+          begin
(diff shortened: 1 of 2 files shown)


bronaghs:
Looks good @djberg96 
Is there a spec that can be added?

djberg96:
@bronaghs I think I would have to create a new spec file, since `provision_spec.rb` stubs out the response for `create_nic`. I noticed that there's no `provision/cloning_spec.rb`, so I'll have to add one, which will take some time.

bronaghs:
I am wondering if you can update `provision_spec.rb `to stub out `NetworkInterfaceService` and `IpAddressService` within the `create_nic` method?

djberg96:
@bronaghs well, i'm not really testing the overall provisioning, just the method itself. Creating a separate spec file for cloning.rb would probably boost our coverage a bit, too.

djberg96:
@bronaghs I've added some specs. I created a separate `cloning_spec.rb` file that just tests the Cloning module methods in isolation instead of adding more provisioning specs since that's really the behavior we care about in this instance.

simaishi:
`gaprindashvili/yes` ?

djberg96:
Well, target release was 5.9, but it should be fine. @bronaghs ?

simaishi:
gaprindashvili is 5.9!

Issue #177, opened 2017-11-29
Azure Debug logging does not contain timestamps


blomquisg:
Looking at the `azure.log` with Debug logging on, I see:

```
RestClient.get "https://nzaccazupdstasdiagdrh1.table.core.windows.net/WADMetricsPT1MP10DV2S20171116?$filter=PartitionKey%20eq%20%27:002Fsubscriptions:<redacted>resourceGroups:<redacted>providers:002FMicrosoft:002ECompute:002FvirtualMachines:002FAZUWDRH1CM4001%27%20and%20CounterName%20eq%20%27%5CPhysicalDisk(_Total)%5CDisk%20Read%20Bytes/sec%27%20and%20Timestamp%20ge%20datetime%272017-11-21T03:15:00Z%27%20and%20Timestamp%20le%20datetime%272017-11-21T04:12:48Z%27&$select=Timestamp,TIMESTAMP,Average", "Accept"=>"application/json;odata=fullmetadata", "Accept-Encoding"=>"gzip, deflate", "User-Agent"=>"rest-client/2.0.2 (linux-gnu x86_64) ruby/2.3.1p112", "auth-string"=>"true", "auth_string"=>"true", "authorization"=>"SharedKey <redacted>", "content-type"=>"", "x-ms-date"=>"Tue, 21 Nov 2017 04:13:12 GMT", "x-ms-version"=>"2016-05-31"
# => 200 OK | application/json 11558 bytes
RestClient.post "https://management.azure.com/subscriptions/<redacted>/resourceGroups/<redacted>/providers/Microsoft.Storage/storageAccounts/<redacted>/listKeys?api-version=2017-10-01", "", "Accept"=>"application/json", "Accept-Encoding"=>"gzip, deflate", "Authorization"=>"Bearer <redacted>", "Content-Length"=>"0", "Content-Type"=>"application/json", "User-Agent"=>"rest-client/2.0.2 (linux-gnu x86_64) ruby/2.3.1p112"
# => 200 OK | application/json 372 bytes
```

This is useful since it shows the size of the response in bytes.  But, it would be even more helpful if it could also show the timing of each request/response.  Or timestamps for each request/response.

I'm not sure if the logging configuration for `RestClient` can get that.  But, I assume it can, because I believe it should be in the header info.  It's just a matter of determining if `RestClient` surfaces that in their logging.

djberg96:
Hm, I thought our logger did that. I'll look into it.

[...]

PR #203, opened 2018-01-22
[WIP] Add timestamp to logging

djberg96:
Attempts to address https://github.com/ManageIQ/manageiq-providers-azure/issues/177


djberg96:
Fixed via https://github.com/ManageIQ/manageiq/pull/17228

PR #210, opened 2018-02-19
Allow users to choose public or private IP when provisioning multiple VMs

djberg96:
When we introduced https://github.com/ManageIQ/manageiq-providers-azure/pull/172, we defaulted to a private IP for provisioning a single VM. This had the unintended side effect of making multi-vm provisioning private by default, with no way for the user to choose.

This PR addresses that by giving the users two choices - new or none - when selecting an IP during the provisioning process for multiple vm's.

Rather than making specialized PR's based on provider on the UI side, this PR uses custom service subclasses that allow us to govern the UI behavior without modifying any core code.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1531275

I have sanity checked this by provisioning two VM's with "none" selected, and two with "new" selected. The app behaved as expected, i.e. the vm's with "none" selected did not have a public IP attached, while those with "new" selected did.

~~WIP for now until I get some specs added.~~

### DIFF
--- a/app/services/manageiq/providers/azure/dialog_field_visibility_service.rb
+++ b/app/services/manageiq/providers/azure/dialog_field_visibility_service.rb
@@ -1,0 +1,19 @@
+class ManageIQ::Providers::Azure::DialogFieldVisibilityService < ::DialogFieldVisibilityService
+  attr_reader :number_of_vms_visibility_service
+
+  def initialize(
+    auto_placement_visibility_service = AutoPlacementVisibilityService.new,
+    number_of_vms_visibility_service = ManageIQ::Providers::Azure::NumberOfVmsVisibilityService.new,
+    service_template_fields_visibility_service = ServiceTemplateFieldsVisibilityService.new,
+    network_visibility_service = NetworkVisibilityService.new,
+    sysprep_auto_logon_visibility_service = SysprepAutoLogonVisibilityService.new,
+    retirement_visibility_service = RetirementVisibilityService.new,
+    customize_fields_visibility_service = CustomizeFieldsVisibilityService.new,
+    sysprep_custom_spec_visibility_service = SysprepCustomSpecVisibilityService.new,
+    request_type_visibility_service = RequestTypeVisibilityService.new,
+    pxe_iso_visibility_service = PxeIsoVisibilityService.new,
+    linked_clone_visibility_service = LinkedCloneVisibilityService.new
+  )
+    super
+  end
+end
--- a/app/services/manageiq/providers/azure/number_of_vms_visibility_service.rb
+++ b/app/services/manageiq/providers/azure/number_of_vms_visibility_service.rb
@@ -1,0 +1,18 @@
+class ManageIQ::Providers::Azure::NumberOfVmsVisibilityService < ::NumberOfVmsVisibilityService
+  def determine_visibility(number_of_vms, platform)
+    @number_of_vms = number_of_vms
+
+    hash = super
+
+    # For Azure we will show two options in this case - private and public
+    if @number_of_vms > 1
+      hash[:hide].delete(:floating_ip_address)
+    end
+
+    hash
+  end
+
+  def number_of_vms
+    @number_of_vms || 1
+  end
+end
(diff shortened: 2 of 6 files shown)


app/services/manageiq/providers/azure/number_of_vms_visibility_service.rb
```diff
@@ -0,0 +1,18 @@
...
+
+    hash = super
+
+    # For Azure we will show two options in this case - private and public
+    if @number_of_vms > 1
+      hash[:hide].delete(:floating_ip_address)
+    end
+
+    hash
+  end
```
bronaghs: @djberg96 - Where is this method called?

app/services/manageiq/providers/azure/dialog_field_visibility_service.rb
```diff
@@ -0,0 +1,19 @@
...
+    service_template_fields_visibility_service = ServiceTemplateFieldsVisibilityService.new,
+    network_visibility_service = NetworkVisibilityService.new,
+    sysprep_auto_logon_visibility_service = SysprepAutoLogonVisibilityService.new,
+    retirement_visibility_service = RetirementVisibilityService.new,
+    customize_fields_visibility_service = CustomizeFieldsVisibilityService.new,
+    sysprep_custom_spec_visibility_service = SysprepCustomSpecVisibilityService.new,
+    request_type_visibility_service = RequestTypeVisibilityService.new,
+    pxe_iso_visibility_service = PxeIsoVisibilityService.new,
+    linked_clone_visibility_service = LinkedCloneVisibilityService.new
+  )
```
bronaghs: Why does this `initialize` method not look like this?
```
def initialize(
 super 
 number_of_vms_visibility_service  = ManageIQ::Providers::Azure::NumberOfVmsVisibilityService.new
)
```

app/services/manageiq/providers/azure/number_of_vms_visibility_service.rb
djberg96: It's used by the `MiqProvisionVirtWorkflow` class at https://github.com/ManageIQ/manageiq/blob/master/app/models/miq_provision_virt_workflow.rb#L985

app/services/manageiq/providers/azure/dialog_field_visibility_service.rb
djberg96: There's no accessor in the superclass.

https://github.com/ManageIQ/manageiq-ui-classic/blob/master/app/services/dialog_field_visibility_service.rb

IMO there should be, though.

bronaghs: I think you should make a PR to add the accessor to the superclass.

djberg96: @bronaghs will do!

djberg96:
@bronaghs https://github.com/ManageIQ/manageiq-ui-classic/pull/3459

But, I wouldn't let that hold up this PR since I don't know if/when it will be merged, and that would have to be backported as well. I can always refactor later.

bronaghs:
Ok, let’s see if we can get the UI PR merged promptly.

djberg96:
@eclarizio I prefer to maintain our own subclass, since that means we don't have to alter core code, and can make any other future changes we want that are Azure specific.

bronaghs:
I agree, let's get this merged given that we are working towards a tight deadline. But let's remember to make the follow up PR... :-)

simaishi:

```
$ git log -1
commit 87ccf8eda5222b5bedbb4c5dbfa6965949aa082b
Date:   Mon Feb 26 13:11:41 2018 -0500

    Merge pull request #210 from djberg96/multi_vm_provision
    
    Allow users to choose public or private IP when provisioning multiple VMs
    (cherry picked from commit 7637bf79a9166fab84f05c95c78c50bc0dddf3ad)
    
    Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1552303
```

simaishi:
Fine backport details:

```
$ git log -1
commit bcf08d78d352be8d4363624eb6f9f17d7775b35f
Date:   Mon Feb 26 13:11:41 2018 -0500

    Merge pull request #210 from djberg96/multi_vm_provision
    
    Allow users to choose public or private IP when provisioning multiple VMs
    (cherry picked from commit 7637bf79a9166fab84f05c95c78c50bc0dddf3ad)
    
    Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1565342
```

PR #212, opened 2018-02-27
Azure graph refresh

Ladas:
Azure graph refresh + specs

Adding also placeholders for targeted refresh, so we can start to add it.

Partially fixes:
https://bugzilla.redhat.com/show_bug.cgi?id=1487602


Ladas:
@bronaghs @djberg96 the CI is failing with region_spec, was there some change recently that would break it?

app/models/manageiq/providers/azure/inventory/parser/cloud_manager.rb
```diff
@@ -0,0 +1,403 @@
...
+      name = deployment.name
+      uid  = deployment.id
+
+      persister_orchestration_stack = persister.orchestration_stacks.build(
+        :ems_ref                => uid,
+        :name                   => name,
+        :description            => name,
+        :status                 => deployment.properties.provisioning_state,
+        :resource_group         => deployment.resource_group,
+        :orchestration_template => persister.orchestration_templates.lazy_find(uid),
```
Ladas: @bzwei so, the easiest way is to no change orchestration_template association, when it's missing, right? Since stack should not even have blank template

bzwei: A stack can have no template associated. but with target refresh I guess if you don't have `:orchestration_template => nil` the existing association remains. In this case the existing association was created by our code during provisioning, not through refresh.
Anyway I agree that we no longer need to load orchestration_template from VMDB.

Ladas: ok, I will move the association elsewhere, only setting it when it's there

For the orchestration_templates themselves, it looks like we do not delete template at all https://github.com/Ladas/manageiq/blob/f77c75b4659f97be95f8976ef58abe521085c963/app/models/orchestration_template.rb#L37 , so we can just send subset everytime (e.g. just fetching template if stack timestamp changed)

so, this could optimize 2nd refresh a lot

Ladas: I'll do this in another PR, possibly extracting it into setting.

app/models/manageiq/providers/azure/cloud_manager/refresh_parser.rb
```diff
@@ -649,7 +649,7 @@ def parse_image(image)
           :publicly_available => false,
           :hardware           => {
             :bitness  => 64,
-            :guest_os => image.operating_system
+            :guest_os => OperatingSystem.normalize_os_name(image.operating_system)
```
bronaghs: @Ladas - slipping this little change into the old parser?

bronaghs:
@bzwei - can you review the orchestration parts?

app/models/manageiq/providers/azure/cloud_manager/refresh_parser.rb
Ladas: Right, I have specs comparing old and new refresh. For some reason old refresh has guest_os normalized, except for this kind of image.

Ladas:
@bronaghs I will be sending the setting for ```Should there be a setting that enables and disables graph and targeted refresh, the default being 'disabled'?``` in the next PRs

app/models/manageiq/providers/azure/inventory/parser/cloud_manager.rb
bzwei: `OrchestrationStack` has `updated_at` timestamp which is the timestamp of the AR record. Raw template deployment has attribute `deployment.properties.timestamp` which is [The timestamp of the template deployment](https://docs.microsoft.com/en-us/rest/api/resources/deployments/get#deploymentpropertiesextended) according to MS document. We may skip the template update if `updated_at` is newer than `timestamp`.

Ladas: yeah, I was looking at that. I'll do it as last PR (after the targeted refresh) as a configurable optimization in settings (since it will mess up most of the refresh specs, by changing API queries in the 2nd refresh)

bronaghs:
@Ladas - Ok, that works for me.

bronaghs:
@Ladas - This PR is great!

simaishi:
@Ladas @bronaghs  There are conflicts on backporting this to gaprindashvili:

```
	both modified:   spec/models/manageiq/providers/azure/cloud_manager/refresher_spec.rb
	both modified:   spec/vcr_cassettes/manageiq/providers/azure/cloud_manager/refresher.yml
```

I can resolve the conflict on refresher_spec.rb, but not sure what need to be done for VCR. Also, I assume other related PRs will have the similar issues, as they all touch VCRs. I'm thinking it might be easier if you create a PR for Gaprindashvili, combining all those PRs. What do you think?

simaishi:
Backported to Gaprindashvili via https://github.com/ManageIQ/manageiq-providers-azure/pull/227

PR #213, opened 2018-02-27
Azure refresh comparison specs

tumido:
Spec: refreshed inventory comparison for Cartesian product of refresh strategies.

Based on: https://github.com/ManageIQ/manageiq-providers-azure/pull/217

@miq-bot add_label enhancement, gaprindashvili/yes, test

cc @Ladas


spec/models/manageiq/providers/azure/cloud_manager/refresher_spec.rb
```diff
@@ -175,6 +182,34 @@ def setup_ems_and_cassette(refresh_settings)
         end
       end
 
+      it "is consistent with old refresh" do
```
Ladas: lets do also other spec turning this around, first graph refresh, then old refresh

spec/models/manageiq/providers/azure/cloud_manager/refresher_spec.rb
```diff
@@ -175,6 +182,34 @@ def setup_ems_and_cassette(refresh_settings)
...
+        inventory_before = serialize_inventory
+        setup_ems_and_cassette(refresh_settings)
+        inventory_after = serialize_inventory
+
+        MODELS.each do |model|
+          expect(inventory_after[model].count).to eq inventory_before[model].count
+
+          inventory_after[model].each do |item|
+            matching_item = inventory_before[model].detect { |i| i["id"] == item["id"] }
+            expect(item).to eq(matching_item), "expected the same inventory for #{model}"
```
Ladas: for the power states, which refresh has it nil, the old refresh?

tumido: Based on the [rspec docs](https://relishapp.com/rspec/rspec-expectations/docs/diffing) the diffs are in a format:
```
- expected
+ actual
```
So in this particular case it means the nil is the expected value of `matching_item` hash (from the old refresh), and the set power state is the actual value (produced by graph refresh).

I know the diff format is not much descriptive and easy to follow, I'll do something about it.

spec/models/manageiq/providers/azure/cloud_manager/refresher_spec.rb
tumido: Sure thing. Will do.

spec/models/manageiq/providers/azure/cloud_manager/refresher_spec.rb
Ladas: hm, interesting, I think you can keep the format, I just asked to get faster understanding :-)

could you print what Vms differ? If all of them, or just some.

tumido: It would fail the expectation each time the entry differs, so if we're seeing it only once, it means only one entry is changed.

spec/models/manageiq/providers/azure/cloud_manager/refresher_spec.rb
tumido: Done

tumido:
Closing in favor of #217, since my changes should be introduced there. :running_man:

PR #219, opened 2018-03-06
Azure graph refresh event target parser

Ladas:
Azure graph refresh event target parser + specs. EventTargetParser is invoked by Automate event handler and parses targets from events, that will be queued for the refresh.

Depends on:
- [x]  https://github.com/ManageIQ/manageiq-providers-azure/pull/217

Partially fixes:
https://bugzilla.redhat.com/show_bug.cgi?id=1487602

PR #221, opened 2018-03-07
Update targeted refresh filters

djberg96:
This updates the targeted refresh filters for most of the resources. For most resources it will gather information for individual resources until the number or resources hits a certain threshold. If that threshold is hit, it will use the current resource collection (i.e. a single request + filter) instead of gathering information for resources one at a time.

You will need to cherry pick the target_collection.rb file out of this PR.


Ladas:
merged as part of https://github.com/ManageIQ/manageiq-providers-azure/pull/219

PR #222, opened 2018-03-13
Azure events targeted

tumido:
1. Add additional event parsing for targeted refresh.
2. Update specs - refresh removal to achieve more unit-like scenario
3. Remove event cartridges, which are no longer needed
4. Add new JSONs for new events

Related: https://github.com/ManageIQ/manageiq-content/pull/261

Partially fixes https://bugzilla.redhat.com/show_bug.cgi?id=1487602

cc @Ladas


app/models/manageiq/providers/azure/cloud_manager/event_target_parser.rb
```diff
@@ -31,30 +31,40 @@ def parse_event_target(target_collection, event_data)
...
                       :security_groups
                     when "Microsoft.Network/networkInterfaces"
                       :network_ports
-                    when "Microsoft.Compute/virtualMachines"
+                    when "Microsoft.Compute/virtualMachines", "Microsoft.Authorization/locks"
                       :vms
                     when "Microsoft.Network/loadBalancers"
                       :load_balancers
-                    when "Microsoft.Network/publicIPAddresses"
+                    when /Microsoft.Network\/publicI[Pp]Addresses/
```
djberg96: You can just use the "i" switch instead of fiddling with character classes. And you may want to escape a literal dot between "Microsoft" and "Network", though it's not a huge deal here.

bronaghs:
@tumido - LGTM. I will merge once you've had a chance to respond to Dan's comment.

app/models/manageiq/providers/azure/cloud_manager/event_target_parser.rb
tumido: Agreed with the dot. Though I'm not sure about the case-insensitive matching - It's a really annoying thingy this Microsoft miracle does. Really. In some cases, for some events (which fortunately we're not using - the commented out `deployments` events), this difference between lowercase and camelcase means also a difference in behavior - so our countermeasures are sometimes counterproductive. It's really a mess.

So I would rather prefer to stay with this precise matching which says: "Only this letter can be uppercase or lowercase and all other text must be precisely this, because here we're sure how it behaves." Or what do you thing?

djberg96: @tumido Where and why do we rely on the case of the resource type? That seems really strange. It shouldn't matter.

Ladas: @djberg96 So that is my original code, we do exact matches in the DB, so it's best to keep everything case sensitive. We transform this to ems_ref to the same format later, otherwise the targeted refresh would not work.

Btw. this is Azure's mess, they have resource_types documented right, but in events it's incorrect. Mostly in events of nested types.

djberg96:
The "pp" it's picking up is just the regex, not a debug print, so we can ignore that.

tumido:

simaishi:

```
$ git log -1
commit dc79a5b1c7bec0a3cc7e997f3c64822e7c8e3751
Date:   Wed Mar 14 10:06:21 2018 -0600

    Merge pull request #222 from tumido/azure_events_targeted
    
    Azure events targeted
    (cherry picked from commit 257d779bc547825dc734093a416943e8033e25b0)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1558078
```

PR #220, opened 2018-03-06
Optimize API collections

Ladas:
Optimize fetching of stack templates and resources. We will fetch templates and resources only if stack's timestamp has changed. Then add optimizations of storage entities and network entities API collection optimizations.

Partially fixes https://bugzilla.redhat.com/show_bug.cgi?id=1487602
Depends on:
- [x] https://github.com/ManageIQ/manageiq-providers-azure/pull/219

Numbers:
the speed improvement for full refresh is 2200s vs 120s now (with private images included)
the targeted refresh of one Vm takes 15-20s, before it was around 40s

refresh of all Vms at once (targeted refresh) ```EmsRefresh.refresh(Vm.all.to_a)``` takes 81s for 82 VMs, it does a lot of n+1 requests, but we have it configurable threshold to fetch all instead of by 1

PR #224, opened 2018-03-15
Add router collection to refresh parser

djberg96:
This adds router support to the standard refresh parser. It collects router and route information, and sets subnet associations, if any.

~~WIP for now because I'll need to add some inventory for testing, which means regenerating the cassettes, and I want to wait until #220 is merged.~~

Addresses https://bugzilla.redhat.com/show_bug.cgi?id=1550605


djberg96:
@Ladas @tumido I'm having a hard time with the specs for this. Even just attempting to regenerate the refresher specs with the updated counts, it's still failing.

Do I have to cover the targeted side as well? I was originally going to do that in a separate PR, but if it's required now, please let me know.

Try running `bundle exec rake app:manageiq:providers:azure:spec:regenerate:refresher` to see what I'm seeing.

Ladas:
@djberg96 hello, at minimum, you will need to do both old and graph refresh. Since there are specs comparing those are the same. Targeted refresh should be ok in another PR, it should not fail any targeted spec, you will need to add specs for refreshing router target.

app/models/manageiq/providers/azure/inventory/parser/network_manager.rb
```diff
@@ -86,10 +86,33 @@ def cloud_subnets(persister_cloud_networks, cloud_network)
         :cidr              => subnet.properties.address_prefix,
         :cloud_network     => persister_cloud_networks,
         :availability_zone => persister.availability_zones.lazy_find('default'),
+        :network_router    => persister.network_routers.lazy_find(subnet.properties.try(:route_table).try(:id))
       )
     end
   end
 
+  def network_routers
+    collect.network_routers.each do |router|
```
Ladas: this will probably fail ```s/collect/collector/```

djberg96: whoops, thanks!

djberg96:
@Ladas I think I'm still missing something, because it complains that the counts are off no matter what I set them too.

What did I miss?

app/models/manageiq/providers/azure/inventory/collector/network_manager.rb
```diff
@@ -10,4 +10,8 @@ def security_groups
   def load_balancers
     @load_balancers ||= gather_data_for_this_region(@lbs)
   end
+
+  def network_routers
+    @network_routers ||= gather_data_for_this_region(@rts)
```
Ladas: maybe lets not memoize this, if we don't have to

Ladas:
@djberg96 can you push the new VCRs so I can see the result here?

djberg96:
@Ladas alright, pushed them up

app/models/manageiq/providers/azure/inventory/parser/network_manager.rb
```diff
@@ -86,10 +86,33 @@ def cloud_subnets(persister_cloud_networks, cloud_network)
         :cidr              => subnet.properties.address_prefix,
         :cloud_network     => persister_cloud_networks,
         :availability_zone => persister.availability_zones.lazy_find('default'),
+        :network_router    => persister.network_routers.lazy_find(subnet.properties.try(:route_table).try(:id))
       )
     end
   end
 
+  def network_routers
```
Ladas: you need to call this method above if ```def parse``` otherwise it's not included

Ladas:
@djberg96 hm, so you will at least need to define :network_router in TargetCollection persister, since parser touches it as a crosslink. Then after including the method in parse as commented above, you should be able to pass the specs. You will need to change the counts to match the current VCR env of course

app/models/manageiq/providers/azure/inventory/parser/network_manager.rb
djberg96: Missed that, thanks.

djberg96:
@Ladas Not sure what you mean by "parser touches it as a crosslink", but I'll add it.

Ladas:
@djberg96 right, for targeted refresh, we refresh subnet that does lazy_find to network_router. So at least defining it for TargetCollection persister, targeted refresh of a subnet will be able to load the lazy link from the DB. Of course not defining it fails in targeted refresh because we do ```persister.network_router.lazy_find```

djberg96:
@Ladas Hm, made the updates, still failing. Help wanted.

app/models/manageiq/providers/azure/inventory/collector/network_manager.rb
```diff
@@ -10,4 +10,8 @@ def security_groups
   def load_balancers
     @load_balancers ||= gather_data_for_this_region(@lbs)
   end
+
+  def network_routers
```
Ladas: if we want to ignore targeted refresh, in Collector::TargetCollection, we should have

```
def network_routers
  [] # TODO add targeted refresh
end
```

then the specs should pass, otherwise it calls the method in the Collector super class

djberg96:
@Ladas ok, ready to go.

simaishi:
@djberg96 There is a conflict backporting to Gaprindashvili because https://github.com/ManageIQ/manageiq-providers-azure/pull/228 isn't in G-branch. Should I backport that PR too, or ignore change for `lib/tasks/azure.rake`?

djberg96:
@simaishi Please backport #228.

simaishi:

```
$ git log -1
commit e6d35af9cf5175f1c5072ea424dc75c8a6c53151
Author: Ladislav Smola <lsmola@redhat.com>
Date:   Tue Mar 27 14:46:11 2018 +0200

    Merge pull request #224 from djberg96/routes
    
    Add router collection to refresh parser
    (cherry picked from commit d14c13241c5250c8dab312ca93fbcbb294a36437)
    
    Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1561085
```

PR #227, opened 2018-03-19
[Gaprindashvili] azure graph refresh

Ladas:
Azure graph refresh + specs

Adding also placeholders for targeted refresh, so we can start to add it.

Partially fixes:
https://bugzilla.redhat.com/show_bug.cgi?id=1487602


Ladas:
@djberg96 ready for review, the last 2 commits are extra, I've bumped the API version, so I can use the upstream VCRs. And 1 difference in graph refresh API, for g-release, that affected LB health checks

Ladas:
@simaishi after @djberg96 reviews, you can proceed with backporting other in https://bugzilla.redhat.com/show_bug.cgi?id=1487602#c18 , locally, it doesn't throw any more conflicts

simaishi:
@Ladas OK, will do!

Ladas:
@simaishi we will need this backported first https://github.com/ManageIQ/manageiq-providers-azure/pull/230 , then the specs here should pass

djberg96:
@Ladas Oh, I thought we weren't going to touch the api-versions in the config file since we merged #230

PR #229, opened 2018-03-20
Azure labels and tag mapping support for a new refresh

AlexanderZagaynov:
BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1493041
related old refresh type PR: https://github.com/ManageIQ/manageiq-providers-azure/pull/231
related core repo PR: https://github.com/ManageIQ/manageiq/pull/17212
related UI repo PR: https://github.com/ManageIQ/manageiq-ui-classic/pull/3697

PR #231, opened 2018-03-27
Azure labels and tags support for old refresh mech

AlexanderZagaynov:
https://bugzilla.redhat.com/show_bug.cgi?id=1568806

new refresh mech: https://github.com/ManageIQ/manageiq-providers-azure/pull/229
related core repo PR: https://github.com/ManageIQ/manageiq/pull/17212
related UI repo PR: https://github.com/ManageIQ/manageiq-ui-classic/pull/3697

### DIFF
--- a/app/models/manageiq/providers/azure/cloud_manager/refresh_parser.rb
+++ b/app/models/manageiq/providers/azure/cloud_manager/refresh_parser.rb
@@ -33,6 +33,8 @@ module ManageIQ::Providers
         @template_uris     = {} # templates need to be download
         @template_refs     = {} # templates need to be retrieved from VMDB
         @template_directs  = {} # templates contents already got by API
+        @tag_mapper        = ContainerLabelTagMapping.mapper
+        @data[:tag_mapper] = @tag_mapper
       end
 
       def ems_inv_to_hashes
@@ -266,6 +268,8 @@ module ManageIQ::Providers
 
         rg_ems_ref = get_resource_group_ems_ref(instance)
 
+        labels = parse_labels(instance.try(:tags) || {})
+
         new_result = {
           :type                => 'ManageIQ::Providers::Azure::CloudManager::Vm',
           :uid_ems             => uid,
@@ -279,6 +283,8 @@ module ManageIQ::Providers
           :orchestration_stack => @data_index.fetch_path(:orchestration_stacks, @resource_to_stack[uid]),
           :availability_zone   => @data_index.fetch_path(:availability_zones, 'default'),
           :resource_group      => @data_index.fetch_path(:resource_groups, rg_ems_ref),
+          :labels              => labels,
+          :tags                => map_labels('VmAzure', labels),
           :hardware            => {
             :disks    => [], # Filled in later conditionally on flavor
             :networks => hardware_network_info
@@ -676,6 +682,19 @@ module ManageIQ::Providers
           stack.delete(:children)
         end
       end
+
+      def parse_labels(tags)
+        tags.map do |tag_name, tag_value|
+          {
+            :name    => tag_name,
+            :value   => tag_value,
+            :source  => 'azure',
+            :section => 'labels',
+          }
+        end
+      end
+
+      delegate :map_labels, :to => :@tag_mapper
     end
   end
 end
--- a/spec/models/manageiq/providers/azure/cloud_manager/refresher_spec.rb
+++ b/spec/models/manageiq/providers/azure/cloud_manager/refresher_spec.rb
@@ -31,6 +31,11 @@ describe ManageIQ::Providers::Azure::CloudManager::Refresher do
         @managed_data_disk = "miqazure-linux-managed-data-disk"
         @template          = nil
         @avail_zone        = nil
+
+        FactoryGirl.create(:tag_mapping_with_category,
+                           :labeled_resource_type => 'VmAzure',
+                           :label_name            => 'Shutdown',
+                           :category_name         => 'azure:vm:shutdown')
       end
 
       after do
@@ -108,6 +113,9 @@ describe ManageIQ::Providers::Azure::CloudManager::Refresher do
       end
 
       it "will perform a full refresh" do
+        # NOTE: for backporting compatibility, where tag mapping was not implemented for the Graph refresh
+        is_old_refresh = !refresh_settings[:inventory_object_refresh]
+
         2.times do # Run twice to verify that a second run with existing data does not change anything
           setup_ems_and_cassette(refresh_settings)
 
@@ -118,8 +126,8 @@ describe ManageIQ::Providers::Azure::CloudManager::Refresher do
           assert_specific_flavor
           assert_specific_disk
           assert_specific_security_group
-          assert_specific_vm_powered_on
-          assert_specific_vm_powered_off
+          assert_specific_vm_powered_on(is_old_refresh)
+          assert_specific_vm_powered_off(is_old_refresh)
           assert_specific_template
           assert_specific_orchestration_template
           assert_specific_orchestration_stack
--- a/spec/models/manageiq/providers/azure/cloud_manager/azure_refresher_spec_common.rb
+++ b/spec/models/manageiq/providers/azure/cloud_manager/azure_refresher_spec_common.rb
@@ -394,7 +394,7 @@ module AzureRefresherSpecCommon
     expect(vm_subnet.floating_ips.size).to be >= 1
   end
 
-  def assert_specific_vm_powered_on
+  def assert_specific_vm_powered_on(is_old_refresh = false)
     vm = ManageIQ::Providers::Azure::CloudManager::Vm.where(
       :name => @device_name, :raw_power_state => "VM running"
     ).first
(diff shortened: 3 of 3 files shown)


app/models/manageiq/providers/azure/cloud_manager/refresh_parser.rb
```diff
@@ -266,6 +268,8 @@ def parse_instance(instance)
 
         rg_ems_ref = get_resource_group_ems_ref(instance)
 
+        labels = parse_labels(instance.tags)
```
cben: These are new right, you didn't collect them before?
Terminology note:
in container provider we called them "labels" in code and used section `labels` because [that's what Kubernetes calls them](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/).
IIUC in AWS and Azure they original concept is called tags?
It gets confusing if we collect "tags" into "labels" and map those into MIQ "tags" :-(

I propose code following UI.  Are you going to display azure tags in UI, when looking at a VM?  How would you call them?
Even if you don't display them, IIRC mapping add/edit UI currently speaks of "label name", "label values" and that needs to be rethought.

- IMHO "single pane of glass" consistency is not a relevant consideration here, if "label" won't mean anything to a user using MIQ with only Azure.
- I can see 2 things called "tags" being confusing as well, but IMHO "Azure tags" vs "ManageIQ/Cloudforms tags" (or "Smart Management tags" if we must™, ugh) could work.

AlexanderZagaynov: I believe it is better to stay in line with Amazon's code patterns here to ease further DRYing (I'm not going to DRY it right now, because it should be backported).
Maybe with the https://github.com/ManageIQ/manageiq-providers-azure/pull/229 PR (which is not suggested for backporting), or with a fully separate one.
What about the UI: I'm not sure how it should be presented there, so I'm aligning to the Amazon's side again for simplicity.

djberg96:
@Ladas You mentioned on the Amazon PR that the specs wouldn't pass until it was also implemented for graph refresh. I assume that's true here as well, correct?

Ladas:
@djberg96 yes, the specs are testing both old refresh and graph refresh, so we either need to add a condition or implement it in both

JPrause:
@miq-bot add_label blocker

AlexanderZagaynov:
I'm closing this in favor of #229 because of https://github.com/ManageIQ/manageiq-providers-azure/pull/231#issuecomment-378499055

AlexanderZagaynov:
https://travis-ci.org/ManageIQ/manageiq-providers-azure/jobs/364148489#L1938
failing because of https://github.com/ManageIQ/manageiq/pull/17212 not yet merged

https://travis-ci.org/ManageIQ/manageiq-providers-azure/jobs/364148489#L1947
failing because of https://github.com/ManageIQ/manageiq-providers-azure/pull/229/commits/04d3085b46ab29325b73cdd5d47b08720daa2c62 can't be merged (for backporting compatibility)

@Ladas @djberg96 @bronaghs any ideas how to workaround the second issue?

bronaghs:
@AlexanderZagaynov Why can't you update the `vm.custom_attributes.size` and `vm.tags.count `from 0 to 1, why cant that be backported to the Gaprindishvili branch?

AlexanderZagaynov:
@bronaghs other tests will fail (that code piece used for all refresh types)

bronaghs:
sounds like those tests needs updated too then if the count legitimately increased on the provider, no?

spec/models/manageiq/providers/azure/cloud_manager/refresh_parser_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+
+  context 'labels and tags' do
+    let(:vm) do
+      VCR.use_cassette(described_class.name.underscore,
+                       :allow_unused_http_interactions => true) { EmsRefresh.refresh(ems) }
+      ems.vms.find_by(:name => 'az-test')
+    end
+
+    it 'fetches labels' do
+      expect(vm.labels.map { |l| [l.name, l.value] }.to_h).to eq('redhat' => 'brno', 'test' => 'hohoho')
```
bronaghs: @AlexanderZagaynov - thinking of readability, can this be simplified to:

`let (:cust_attr) do
    vm.labels.first
end`

expect(cust_attr.name).to eq('redhat')
expect(cust_attr.value).to eq('brno')

bronaghs:
hi @AlexanderZagaynov - I am curious why you created a new spec and didnt just add these tests to an existing file? The first 23ish lines are duplicates.

spec/models/manageiq/providers/azure/cloud_manager/refresher_spec.rb
```diff
@@ -118,8 +122,8 @@
           assert_specific_flavor
           assert_specific_disk
           assert_specific_security_group
-          assert_specific_vm_powered_on
-          assert_specific_vm_powered_off
+          assert_specific_vm_powered_on(**assertion_options)
+          assert_specific_vm_powered_off(**assertion_options)
           assert_specific_template
```
bronaghs: do you need the double splat when you are passing in 1 argument?

bronaghs:
Looks good @AlexanderZagaynov

AlexanderZagaynov:
@miq-bot add_label "fine/no"

AlexanderZagaynov:
@miq-bot add_label fine/no

JPrause:
@AlexanderZagaynov can you add the gaprindashvili/yes and the fine/yes if this can be backported.

AlexanderZagaynov:
@bronaghs specs are not failing now

spec/models/manageiq/providers/azure/cloud_manager/refresher_spec.rb
```diff
@@ -108,6 +113,9 @@
       end
 
       it "will perform a full refresh" do
+        # NOTE: for backporting compatibility, where tagging does not implemented for the Graph refresh
```
martinpovolny: The comment does not sound English ;-)
I am not sure how to fix it. Please, ask some native speaker.

JPrause: Not sure of the context but I'd suggest either "is not implemented" or "can not be implemented".

JPrause:
@AlexanderZagaynov please add the gaprindashvili/yes label if this can be backported. Can you also verify the fine/no label.

spec/models/manageiq/providers/azure/cloud_manager/refresher_spec.rb
cben: Perhaps "where tag mapping was not implemented"

Ladas: lets just fix this and then I think its ready

spec/models/manageiq/providers/azure/cloud_manager/refresher_spec.rb
```diff
@@ -31,6 +31,11 @@
         @managed_data_disk = "miqazure-linux-managed-data-disk"
         @template          = nil
         @avail_zone        = nil
+
+        FactoryGirl.create(:tag_mapping_with_category,
+                           :labeled_resource_type => 'VmAzure',
```
Ladas: the class should be ``` ManageIQ::Providers::Azure::CloudManager::Vm```

app/models/manageiq/providers/azure/cloud_manager/refresh_parser.rb
```diff
@@ -279,6 +283,8 @@ def parse_instance(instance)
           :orchestration_stack => @data_index.fetch_path(:orchestration_stacks, @resource_to_stack[uid]),
           :availability_zone   => @data_index.fetch_path(:availability_zones, 'default'),
           :resource_group      => @data_index.fetch_path(:resource_groups, rg_ems_ref),
+          :labels              => labels,
+          :tags                => map_labels('VmAzure', labels),
```
Ladas: the class should be ``` ManageIQ::Providers::Azure::CloudManager::Vm```

AlexanderZagaynov: that is handled there: https://github.com/ManageIQ/manageiq-ui-classic/pull/3697/files#diff-95b50e997ace14a4e385c9a176eea25fR85
and we discussed it with @cben a lot, and agreed on current implementation, please don't make us starting that over again :)

spec/models/manageiq/providers/azure/cloud_manager/refresher_spec.rb
AlexanderZagaynov: done

app/models/manageiq/providers/azure/cloud_manager/refresh_parser.rb
cben: @Ladas this is not supposed to be a class name.  it's just a string that matches what UI puts in ContainerLabelTagMappings (if constrained to one type).  you can see a list of currently used values in Alexander's link ^^ in *keys* of MAPPABLE_ENTITIES const.  see for example `"Image"` which is not a class name.

Yes, this arrangement could be improved as a whole (with a migration — out of scope here), but as it is, adding 1 fully qualified model name to that set would only confuse matters IMO.

cben:
locally, cherry-picks to gaprindashvili cleanly, and passes tests.
@miq-bot add-label gaprindashvili/yes, fine/conflict

AlexanderZagaynov:
There is a separate PR for **Fine**: https://github.com/ManageIQ/manageiq-providers-azure/pull/242

AlexanderZagaynov:
@miq-bot remove-label fine/conflict

simaishi:
Backported to Fine via https://github.com/ManageIQ/manageiq-providers-azure/pull/242

simaishi:

```
$ git log -1
commit ea578dbf0e2eb76dc34aca04ba18fb2fd59684ce
Date:   Wed Apr 18 13:48:32 2018 +0200

    Merge pull request #231 from AlexanderZagaynov/BZ-1493041-old-refresh
    
    Azure labels and tags support for old refresh mech
    (cherry picked from commit 3014aee644b6280207332f7ac9a108b4c618b800)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1568807
```

PR #233, opened 2018-04-04
Ignore deployment events

djberg96:
No action needs to be executed when the events `deployments_exportTemplate_BeginRequest` and `deployments_exportTemplate_EndRequest` are received, so we should blacklist them.

Addresses https://bugzilla.redhat.com/show_bug.cgi?id=1561646

~~WIP for now as it doesn't seem to actually be working.~~

Also addresses the fact that the settings file was being ignored, as the ignored events were hard coded into the cloud manager.

### DIFF
--- a/app/models/manageiq/providers/azure/cloud_manager.rb
+++ b/app/models/manageiq/providers/azure/cloud_manager.rb
@@ -59,10 +59,7 @@ class ManageIQ::Providers::Azure::CloudManager < ManageIQ::Providers::CloudManag
   end
 
   def self.default_blacklisted_event_names
-    %w(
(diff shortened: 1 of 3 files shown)


Ladas:
@djberg96 the code filtering the events is here https://github.com/ManageIQ/manageiq-providers-azure/blob/master/app/models/manageiq/providers/azure/cloud_manager/event_catcher/runner.rb#L47

djberg96:
@Ladas Yep, I see that. The problem was that the cloud manager had the values hard-coded instead of reading from the config file.

Ladas:
cc @bronaghs marking as performance bug, it should be beneficial to backport this as far as we can

PR #234, opened 2018-04-06
Fix specs for targeted api collection threshold

Ladas:
Because of a wrong merge, the targeted_api_collection_threshold with
value 0 was not tested, so we were not testing targeted refresh that
fetches the whole collection, instead of by 1 record.


spec/models/manageiq/providers/azure/cloud_manager/targeted_refresher_spec.rb
```diff
@@ -10,7 +10,7 @@
         context "with additional settings #{additional_settings}" do
           before(:each) do
             @refresh_settings = refresh_settings.merge(:allow_targeted_refresh => true)
-            @refresh_settings.merge(additional_settings)
```
Ladas: @bronaghs @djberg96 this was the issue, the settings were not merged, so we were not testing the treshold that switches to collect all API records and filter in ruby vs. collect records by 1 from the API.

Ladas:
@bronaghs we should backport it, so we have it tested in g-release, but not as a blocker for this release

djberg96:
@Ladas You've got a couple failing specs.

Ladas:
@djberg96 uf, finally I made it pass. It was passing locally again. The issue is the resource_group ems_ref has .downcase on it. But then the obfuscation is adding the upcase SUBSCRIPTION_ID, so that was breaking the CI specs.

djberg96:
👍

PR #235, opened 2018-04-09
Change saver_strategy value to String

Fryguy:
Symbols in configuration values are problematic because Symbols do not
roundtrip through JSON.  Since the API now exposes Settings, it's not
possible to set Symbols as values.  The saver_strategy was fixed in
https://github.com/ManageIQ/manageiq/pull/17168 to support String
values, and the next step is to remove all Symbols from the Settings.

https://github.com/ManageIQ/manageiq/issues/17201
https://bugzilla.redhat.com/show_bug.cgi?id=1558031

@Ladas Please review


cben:
Can this be gaprindashvili/yes?
(to be backported after https://github.com/ManageIQ/manageiq/pull/17168)

PR #236, opened 2018-04-09
Convert api-version strings to actual strings

djberg96:
Followup to https://github.com/ManageIQ/manageiq/issues/17201 and https://github.com/ManageIQ/manageiq-providers-azure/pull/235. Eliminates lots of unnecessary `to_s` calls as well.


djberg96:
@Fryguy Look alright?

PR #240, opened 2018-04-13
Ensure image name is clean

djberg96:
The `build_image_name` method will create a leading "./" if the dirname is relative. This is because `File.dirname("foo")` returns "." on relative path names.

~~WIP for now until I can see if there's a nicer way without involving Pathname.~~

I didn't see a better way, so Pathname it is.

See also https://github.com/ManageIQ/manageiq-schema/pull/198

### DIFF
--- a/app/models/manageiq/providers/azure/inventory/parser/cloud_manager.rb
+++ b/app/models/manageiq/providers/azure/inventory/parser/cloud_manager.rb
@@ -449,7 +449,8 @@ class ManageIQ::Providers::Azure::Inventory::Parser::CloudManager < ManageIQ::Pr
 
   def build_image_name(image)
     # Strip the .vhd and Azure GUID extension, but retain path and base name.
-    File.join(File.dirname(image.name), File.basename(File.basename(image.name, '.*'), '.*'))
+    path = File.join(File.dirname(image.name), File.basename(File.basename(image.name, '.*'), '.*'))
+    Pathname.new(path).cleanpath.to_s
   end
 
   def build_image_description(image)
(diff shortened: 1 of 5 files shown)


djberg96:
@bzwei What do you think? Is there a nicer way?

Also, this will require a migration.

bzwei:
@djberg96 what's the problem - is it a bug or simply a presentation issue? Normally "." in a path does not cause a problem, therefore a migration is not needed.

djberg96:
@bzwei Presentation issue. Seems weird to have "./foo" as an image name. I'll double check, but without the migration users would end up with "./foo" and "foo" as separate records, wouldn't they?

djberg96:
@bzwei I added a migration in the schema repo as well.

bzwei:
consider to add a spec test?

djberg96:
@bzwei Alright, I guess I added them for the migration but not here. Will add them.

djberg96:
@bzwei Ok, added specs. Also, I moved it into the helper methods module so that I could more easily test it, and possibly re-use it in the targeted refresh code eventually.

app/models/manageiq/providers/azure/inventory/parser/cloud_manager.rb
```diff
@@ -449,7 +449,8 @@ def resource_status_message(resource)
 
   def build_image_name(image)
     # Strip the .vhd and Azure GUID extension, but retain path and base name.
-    File.join(File.dirname(image.name), File.basename(File.basename(image.name, '.*'), '.*'))
+    path = File.join(File.dirname(image.name), File.basename(File.basename(image.name, '.*'), '.*'))
```
bzwei: can you use `build_image_name` method in `refresh_helper_methods`?

djberg96: I'd have to include the module. I don't think it will be an issue, but I'll give it a try.

PR #244, opened 2018-04-20
Bump to ruby 2.3.6

jrafanie:
Bump to ruby 2.3.6

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: ruby
 rvm:
-- '2.3.1'
-- '2.4.2'
+- 2.3.6
+- 2.4.2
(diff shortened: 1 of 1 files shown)


djberg96:
It doesn't need to be quoted?

jrafanie:
> It doesn't need to be quoted?

```
irb(main):002:0> YAML.dump "2.3.1"
=> "--- 2.3.1\n...\n"
```

No. I generated this by yaml dumping [here](https://github.com/ManageIQ/manageiq-release/pull/60) in the PR blaster outer script.  It would be nice to follow the format that YAML dumps so we can blast the same update to `.travis.yml` on several repos at once.

PR #245, opened 2018-04-23
Forward slash all the things

djberg96:
When we first started adding Azure provider support, we made the decision to incorporate backslashes. This proved to be unwise, as the Azure REST API uses forward slashes everywhere. The net result has been the proliferation of the `resource_uid` method and confusion in general. Plus, we've ended up with ems_ref's that are a mix of forward slashes and backslashes. It's just dumb.

This PR attempts to eliminate that issue by using forward slashes eveywhere.

Note that there is one test failure caused by a factory in the core repo that will need to be updated. In addition, a separate data migration will need to be implemented in parallel with this PR.

### DIFF
--- a/app/models/manageiq/providers/azure/cloud_manager/event_target_parser.rb
+++ b/app/models/manageiq/providers/azure/cloud_manager/event_target_parser.rb
@@ -113,7 +113,7 @@ class ManageIQ::Providers::Azure::CloudManager::EventTargetParser
   def resource_id_for_instance_id(id)
     return nil unless id
     _, _, guid, _, resource_group, _, type, sub_type, name = id.split("/")
-    File.join(guid, resource_group.downcase, "#{type.downcase}/#{sub_type.downcase}", name)
+    File.join(guid, resource_group.downcase, type.downcase, sub_type.downcase, name)
   end
 
   def resource_id_for_cloud_networks(id)
--- a/app/models/manageiq/providers/azure/cloud_manager/event_target_parser.rb
+++ b/app/models/manageiq/providers/azure/cloud_manager/event_target_parser.rb
@@ -113,10 +113,7 @@ class ManageIQ::Providers::Azure::CloudManager::EventTargetParser
   def resource_id_for_instance_id(id)
     return nil unless id
     _, _, guid, _, resource_group, _, type, sub_type, name = id.split("/")
-    resource_uid(guid,
-                 resource_group.downcase,
-                 "#{type.downcase}/#{sub_type.downcase}",
-                 name)
+    File.join(guid, resource_group.downcase, "#{type.downcase}/#{sub_type.downcase}", name)
   end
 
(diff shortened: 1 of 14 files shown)


djberg96:
@Ladas Yep, I'll try and put that together today.

https://github.com/ManageIQ/manageiq/pull/17337

app/models/manageiq/providers/azure/cloud_manager/event_target_parser.rb
```diff
@@ -113,10 +113,7 @@ def resource_id_for_stack_id(id)
   def resource_id_for_instance_id(id)
     return nil unless id
     _, _, guid, _, resource_group, _, type, sub_type, name = id.split("/")
-    resource_uid(guid,
-                 resource_group.downcase,
-                 "#{type.downcase}/#{sub_type.downcase}",
-                 name)
+    File.join(guid, resource_group.downcase, "#{type.downcase}/#{sub_type.downcase}", name)
```
Ladas: we could probably replace ```"#{type.downcase}/#{sub_type.downcase}"``` with ```type.downcase, sub_type.downcase``` now?

app/models/manageiq/providers/azure/event_catcher_mixin.rb
```diff
@@ -30,6 +30,6 @@ def parse_vm_ref(event)
     object_class,
     object_name = event["resourceId"].split("/")
 
-    [subscription_id, resource_group.downcase, "#{provider.downcase}\/#{object_class.downcase}", object_name].join("\\")
+    [subscription_id, resource_group.downcase, provider.downcase, object_class.downcase, object_name].join("/")
```
Ladas: should we also use File.join, to be consistent?

app/models/manageiq/providers/azure/inventory/collector/target_collection.rb
```diff
@@ -594,19 +594,11 @@ def add_simple_target!(association, ems_ref)
...
   def resource_id_for_instance_id(id)
     # TODO(lsmola) we really need to get rid of the building our own emf_ref, it makes crosslinking impossible, parsing
     # the id string like this is suboptimal
     return nil unless id
     _, _, guid, _, resource_group, _, type, sub_type, name = id.split("/")
-    resource_uid(guid,
-                 resource_group.downcase,
-                 "#{type.downcase}/#{sub_type.downcase}",
-                 name)
+    File.join(guid, resource_group.downcase, "#{type.downcase}/#{sub_type.downcase}", name)
```
Ladas: ```type.downcase, sub_type.downcase```

djberg96:
@Ladas good catch. I'll make those updates, and run it through rubocop.

djberg96:
@Ladas Let's do a separate PR after this for rubocop cleanup, since it's much more than just these files.

Fryguy:
Merged schema and the core change, so rekicked travis

PR #249, opened 2018-05-03
Rubocop cleanup

djberg96:
Another round of rubocop cleanup, this fixes 33 of the remaining 70 rubocop issues. The remaining issues are either invalid or would require additional refactoring.

### DIFF
--- a/app/models/manageiq/providers/azure/regions.rb
+++ b/app/models/manageiq/providers/azure/regions.rb
@@ -139,8 +139,8 @@ module ManageIQ
       # disabled regions.
       #
       def self.regions
-        additional_regions = Hash(Settings.ems.ems_azure.try!(:additional_regions)).stringify_keys
-        disabled_regions   = Array(Settings.ems.ems_azure.try!(:disabled_regions))
+        additional_regions = Hash(Settings.ems.ems_azure&.additional_regions).stringify_keys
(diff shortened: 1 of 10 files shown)


app/models/manageiq/providers/azure/regions.rb
```diff
@@ -139,8 +139,8 @@ module Regions
       # disabled regions.
       #
       def self.regions
-        additional_regions = Hash(Settings.ems.ems_azure.try!(:additional_regions)).stringify_keys
-        disabled_regions   = Array(Settings.ems.ems_azure.try!(:disabled_regions))
```
bronaghs: Dan and I are not sure what the point of having the `!` here was. @agrare @Ladas - do you know? I want to make sure we are not eliminating some functionality through this PR.

djberg96: The semantics for `&.` and `try!` are the same, so this won't break anything, but I guess the question is why we're using `try!` instead of `try` in the first place.

agrare: This looks like it was copied and pasted from my code in the amazon repo which I coped and pasted from Marcel haha
We aren't relying on the difference between `.try!` and `try`, `&.` should be fine here.

PR #255, opened 2018-05-14
Archive VM with no status

Ladas:
Archive VM with no status, which should solve that VMs are not
sometimes archived by targeted refresh. We receive event that VM
was deleted, but when doing API query, the Vm is still returned.
When fetching Vm's state, we do get uknown state (azure returns 404)

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1575773

### DIFF
--- a/app/models/manageiq/providers/azure/inventory/parser/cloud_manager.rb
+++ b/app/models/manageiq/providers/azure/inventory/parser/cloud_manager.rb
@@ -70,12 +70,15 @@ class ManageIQ::Providers::Azure::Inventory::Parser::CloudManager < ManageIQ::Pr
 
       rg_ems_ref = collector.get_resource_group_ems_ref(instance)
 
+      # We want to archive VMs with no status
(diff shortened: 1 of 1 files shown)


app/models/manageiq/providers/azure/inventory/parser/cloud_manager.rb
```diff
@@ -70,12 +70,15 @@ def instances
 
       rg_ems_ref = collector.get_resource_group_ems_ref(instance)
 
+      # We want to archive VMs with no status
+      next if (status = collector.power_status(instance)).blank?
```
bzwei: do you want to skip all the rest of code or simply want to set the `raw_power_state` to be `nil`?

Ladas: point is to skip ```persister.vms.build``` which will cause the archival of the existing Vm

PR #257, opened 2018-05-15
Handle providers that may not support managed images or disks

djberg96:
At least one region (US Gov Iowa) doe not currently support managed images or disks. As a result, the refresh parser will blow up when it attempts to collect that information. This PR adds a check to see if the provider supports managed images and disks before attempting to collect that information. If not it simply returns an empty array.

https://azure.microsoft.com/en-us/global-infrastructure/services/

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1566600

Although the BZ listed above was originally written for Azure Stack, it applies generally.


djberg96:
@bsquizz I need to add some specs, but this should work for Azure stack.

bronaghs:
@djberg96 Why did this call for all the cassettes to be regenerated?

djberg96:
@bronaghs Because the `supported?` check is actually making a request to the Azure REST API, so that has to be recorded.

bronaghs:
Spoke to Dan offline, the change in API version led to the regeneration of many new cassettes. This widespread update did not break any tests.

bronaghs:
@djberg96 - can you set the backport labels?

PR #264, opened 2018-06-11
Fix root disk size, swap disk size

djberg96:
We are mistakenly storing kilobytes instead of bytes for `root_disk_size` and `swap_disk_size` at the moment, which also affects the hardware information. Ultimately, this is causing automation issues where disk quotas have been set.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1564495


agrare:
FTR here is where we are expecting `root_disk_size` to be in bytes, https://github.com/ManageIQ/manageiq/blob/master/app/models/flavor.rb#L38

PR #267, opened 2018-06-20
Set parent for VM's where possible

djberg96:
This PR sets the parent image for virtual machines in Azure. Previously, this was only getting set via provisioning within the tool.

Note that this does not currently support marketplace images. Setting this is not impossible, but will require a change in how we collect and/or store marketplace image data. This is because Azure always sets the version to "latest" for a VM, whereas there can be many different versions for the same markeplace image, and more are added over time.

Addresses https://bugzilla.redhat.com/show_bug.cgi?id=1478889

~~WIP for now until targeted refresh is fixed and the specs pass.~~

### DIFF
--- a/app/models/manageiq/providers/azure/inventory/parser/cloud_manager.rb
+++ b/app/models/manageiq/providers/azure/inventory/parser/cloud_manager.rb
@@ -71,6 +71,7 @@ class ManageIQ::Providers::Azure::Inventory::Parser::CloudManager < ManageIQ::Pr
       series = persister.flavors.find(instance.properties.hardware_profile.vm_size.downcase)
 
       rg_ems_ref = collector.get_resource_group_ems_ref(instance)
+      parent_ref = collector.parent_ems_ref(instance)
 
       # We want to archive VMs with no status
       next if (status = collector.power_status(instance)).blank?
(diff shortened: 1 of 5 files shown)


djberg96:
@Ladas @tumido I need some help here. I'm not sure how to set the parent for targeted refresh.

app/models/manageiq/providers/azure/inventory/parser/cloud_manager.rb
```diff
@@ -83,6 +84,7 @@ def instances
         :raw_power_state     => status,
         :flavor              => series,
         :location            => instance.location,
+        :parent              => persister.vms.lazy_find(parent_ref),
```
Ladas: it looks like the parent is an Image? So this shouldbe correct:
```ruby
:parent              => persister.miq_templates.lazy_find(parent_ref),
```

Ladas: this is supposed to be image used for provisioning of the Vm, right?

djberg96:
@Ladas Ok, made that change. Now there's all sorts of weird failures. :(

app/models/manageiq/providers/azure/inventory/parser/cloud_manager.rb
```diff
@@ -83,6 +84,7 @@ def instances
         :raw_power_state     => status,
         :flavor              => series,
         :location            => instance.location,
+        :parent              => persister.miq_templates.lazy_find(parent_ref),
```
Ladas: it has to be called ```genealogy_parent```, there is after save logic handling this as backup, but we use custom saver saving this

PR #271, opened 2018-06-21
Persister: optimized InventoryCollection definitions

slemrmartin:
Optimizations and simplification of InventoryCollection definitions

**Issue**: https://github.com/ManageIQ/manageiq/issues/17396
- [x] **depends on**: https://github.com/ManageIQ/manageiq/pull/17621

What is possible is moved to core


slemrmartin:
cc @agrare

agrare:
Looks like this is failing on the same spec that master is currently failing on

agrare:
Should be fixed by https://github.com/ManageIQ/manageiq-providers-azure/pull/273

Issue #252, opened 2018-05-07
Azure Stack flavor names are case sensitive


bsquizz:
During Azure Stack testing, I tried to deploy a a private image and hit this error:

```
[----] E, [2018-05-07T12:27:03.581936 #5157:121f114] ERROR -- : Q-task_id([miq_provision_1]) MIQ(ManageIQ::Providers::Azure::CloudManager::Provision#provision_error) [[Azure::Armrest::BadRequestException]: The requested VM size standard_a1 is not available in the current region. The sizes available in the current region are: Basic_A0,Basic_A1,Basic_A2,Basic_A3,Basic_A4,Standard_A0,Standard_A1,Standard_A2,Standard_A3,Standard_A4,Standard_A5,Standard_A6,Standard_A7,Standard_D1,Standard_D2,Standard_D3,Standard_D4,Standard_D11,Standard_D12,Standard_D13,Standard_D14,Standard_D1_v2,Standard_D2_v2,Standard_D3_v2,Standard_D4_v2,Standard_D5_v2,Standard_D11_v2,Standard_D12_v2,Standard_D13_v2,Standard_D14_v2,Standard_DS1,Standard_DS2,Standard_DS3,Standard_DS4,Standard_DS11,Standard_DS12,Standard_DS13,Standard_DS14,Standard_DS1_v2,Standard_DS2_v2,Standard_DS3_v2,Standard_DS4_v2,Standard_DS5_v2,Standard_DS11_v2,Standard_DS12_v2,Standard_DS13_v2,Standard_DS14_v2. Find out more on the available VM sizes in each region at https://aka.ms/azure-regions.] encountered during phase [start_clone_task]
```

Requested vm size was 'standard_a1', and Standard_A1 is in the list. It appears that the flavor name is case sensitive for the Azure Stack APIs.

We could get around this by only storing the uid as `flavor.name.downcase` and keeping the name as-is (see https://github.com/ManageIQ/manageiq-providers-azure/compare/master...bsquizz:flavor_names), however I'm not sure if altering the display name will affect anything existing customers are doing with Azure providers. If that's the case, we may need to store another attribute (i.e. `actual_name`) alongside the downcase'd `name` and have `prepare_for_clone_task` in provision/cloning.rb use the `actual_name` for `vmSize`

djberg96:
[...]

PR #272, opened 2018-06-21
Update series name to work with Azure stack

djberg96:
Apparently flavor names are case sensitive for Azure Stack. This PR keeps the downcased name for the `ems_ref`, but retains the original name for the `name` field.

Addresses https://github.com/ManageIQ/manageiq-providers-azure/issues/252

### DIFF
--- a/app/models/manageiq/providers/azure/cloud_manager/refresh_parser.rb
+++ b/app/models/manageiq/providers/azure/cloud_manager/refresh_parser.rb
@@ -230,18 +230,20 @@ module ManageIQ::Providers
         return uid, new_result
       end
 
-      def parse_series(s)
-        name = uid = s.name.downcase
+      def parse_series(series)
+        uid = series.name.downcase
+
         new_result = {
-          :type           => "ManageIQ::Providers::Azure::CloudManager::Flavor",
+          :type           => ManageIQ::Providers::Azure::CloudManager::Flavor.name,
           :ems_ref        => uid,
-          :name           => name,
-          :cpus           => s.number_of_cores, # where are the virtual CPUs??
-          :cpu_cores      => s.number_of_cores,
-          :memory         => s.memory_in_mb.megabytes,
-          :root_disk_size => s.os_disk_size_in_mb.megabytes,
-          :swap_disk_size => s.resource_disk_size_in_mb.megabytes
+          :name           => series.name,
+          :cpus           => series.number_of_cores,
+          :cpu_cores      => series.number_of_cores,
+          :memory         => series.memory_in_mb.megabytes,
+          :root_disk_size => series.os_disk_size_in_mb.megabytes,
+          :swap_disk_size => series.resource_disk_size_in_mb.megabytes
         }
+
         return uid, new_result
       end
 
--- a/app/models/manageiq/providers/azure/inventory/collector/target_collection.rb
+++ b/app/models/manageiq/providers/azure/inventory/collector/target_collection.rb
@@ -46,13 +46,13 @@ class ManageIQ::Providers::Azure::Inventory::Collector::TargetCollection < Manag
   end
 
   def flavors
-    refs = name_references(:flavors)
+    refs = references(:flavors)
 
     return [] if refs.blank?
     set = Set.new(refs)
 
     collect_inventory_targeted(:series) { @vmm.series(@ems.provider_region) }.select do |flavor|
-      set.include?(flavor.name.downcase)
+      set.include?(flavor.name.downcase) # ems_ref is downcased flavor name
     end
   rescue ::Azure::Armrest::Exception => err
     _log.error("Error Class=#{err.class.name}, Message=#{err.message}")
@@ -521,7 +521,7 @@ class ManageIQ::Providers::Azure::Inventory::Collector::TargetCollection < Manag
   def infer_related_vm_ems_refs_api!
     instances.each do |instance|
       # TODO(lsmola) add API scanning
-      target.add_target(:association => :flavors, :manager_ref => {:name => instance.properties.hardware_profile.vm_size.downcase})
+      target.add_target(:association => :flavors, :manager_ref => {:ems_ref => instance.properties.hardware_profile.vm_size.downcase})
       add_simple_target!(:availability_zones, 'default')
       add_simple_target!(:resource_groups, get_resource_group_ems_ref(instance))
       instance.properties.network_profile.network_interfaces.collect(&:id).each do |network_port_ems_ref|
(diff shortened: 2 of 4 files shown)


app/models/manageiq/providers/azure/cloud_manager/refresh_parser.rb
```diff
@@ -230,17 +230,16 @@ def parse_resource_group(resource_group)
...
+          :type           => ManageIQ::Providers::Azure::CloudManager::Flavor.name,
+          :ems_ref        => series.name.downcase,
+          :name           => series.name,
+          :cpus           => series.number_of_cores,
+          :cpu_cores      => series.number_of_cores,
+          :memory         => series.memory_in_mb.megabytes,
+          :root_disk_size => series.os_disk_size_in_mb.megabytes,
+          :swap_disk_size => series.resource_disk_size_in_mb.megabytes
         }
         return uid, new_result
```
bronaghs: @djberg96 - Where is the UID set?

djberg96: Updated.

app/models/manageiq/providers/azure/cloud_manager/refresh_parser.rb
```diff
@@ -230,18 +230,20 @@ def parse_resource_group(resource_group)
...
-          :type           => "ManageIQ::Providers::Azure::CloudManager::Flavor",
+          :type           => ManageIQ::Providers::Azure::CloudManager::Flavor.name,
           :ems_ref        => uid,
-          :name           => name,
-          :cpus           => s.number_of_cores, # where are the virtual CPUs??
-          :cpu_cores      => s.number_of_cores,
-          :memory         => s.memory_in_mb.megabytes,
-          :root_disk_size => s.os_disk_size_in_mb.megabytes,
-          :swap_disk_size => s.resource_disk_size_in_mb.megabytes
+          :name           => series.name.split('_').map(&:capitalize).join('_'),
```
bronaghs: Why are you not saving the name as it is pulled in from Azure? I would think the name we send back  **_to_** Azure during provisioning should match what we receive _**from**_ Azure in refresh.

djberg96: I thought it might be best to make sure, but I can undo that, as I guess it could backfire.

bronaghs:
@djberg96 - What about target refresh? See [here](https://github.com/ManageIQ/manageiq-providers-azure/blob/master/app/models/manageiq/providers/azure/inventory/collector/target_collection.rb#L48..L60)

bronaghs:
Looks like you need to do a thorough search through the graph refresh code, [here](https://github.com/ManageIQ/manageiq-providers-azure/blob/master/app/models/manageiq/providers/azure/inventory/parser/cloud_manager.rb#L71) is another reference to downcasing the flavor

app/models/manageiq/providers/azure/inventory/collector/target_collection.rb
```diff
@@ -521,7 +521,7 @@ def infer_related_vm_ems_refs_db!
   def infer_related_vm_ems_refs_api!
     instances.each do |instance|
       # TODO(lsmola) add API scanning
-      target.add_target(:association => :flavors, :manager_ref => {:name => instance.properties.hardware_profile.vm_size.downcase})
+      target.add_target(:association => :flavors, :manager_ref => {:name => instance.properties.hardware_profile.vm_size})
```
djberg96: I tried switching this to use `ems_ref` with the downcase'd version, but that seemed to make it unhappy.

djberg96:
@bronaghs @Ladas @tumido At this point I need an extra set of eyes, as I'm not sure why it's failing. The couple places where I left the downcase'd version I left because I think it's looking for the `ems_ref`.

app/models/manageiq/providers/azure/inventory/collector/target_collection.rb
tumido: May that be caused by using `:name` as `:manager_ref`? Shouldn't you rather use `:ems_ref`? I mean like this:

```ruby
target.add_target(:association => :flavors, :manager_ref => {:ems_ref => instance.properties.hardware_profile.vm_size.downcase})
```

Ladas: yeah, looks like flavor has defined ```:manager_ref => [:ems_ref]``` so we should use ems_ref ( this looks like a bug )

Ladas: parser code for graph refresh is here, it worked because ems_ref and name are the same https://github.com/Ladas/manageiq-providers-azure/blob/949a23dd25ab883a87fe5d5b430442d135e9434d/app/models/manageiq/providers/azure/inventory/parser/cloud_manager.rb#L36

app/models/manageiq/providers/azure/inventory/collector/target_collection.rb
```diff
@@ -52,7 +52,7 @@ def flavors
     set = Set.new(refs)
 
     collect_inventory_targeted(:series) { @vmm.series(@ems.provider_region) }.select do |flavor|
-      set.include?(flavor.name.downcase)
```
Ladas: I think this should stay and we should use ```refs = references(:flavors)``` on L49

app/models/manageiq/providers/azure/inventory/collector/target_collection.rb
```diff
@@ -521,7 +521,7 @@ def infer_related_vm_ems_refs_db!
   def infer_related_vm_ems_refs_api!
     instances.each do |instance|
       # TODO(lsmola) add API scanning
-      target.add_target(:association => :flavors, :manager_ref => {:name => instance.properties.hardware_profile.vm_size.downcase})
```
Ladas: again, if ems_ref is downcased, we should keep this

app/models/manageiq/providers/azure/inventory/collector/target_collection.rb
djberg96: Ok, I've made that change.

app/models/manageiq/providers/azure/inventory/collector/target_collection.rb
djberg96: @tumido @Ladas I've updated that to point at the ems_ref instead.

app/models/manageiq/providers/azure/inventory/collector/target_collection.rb
djberg96: @Ladas Ok, restored the downcase, but switched it to use `ems_ref` instead of `name`.

djberg96:
@bronaghs @Ladas @tumido Made some changes, still stuck. Any suggestions?

Ladas:
@djberg96 you have modified only the old refresh parser. So you need to change it also here https://github.com/Ladas/manageiq-providers-azure/blob/b21fe6632e12018688800ed9ba4ff76c0daf6bcb/app/models/manageiq/providers/azure/inventory/parser/cloud_manager.rb#L38

PR #273, opened 2018-06-29
Don't use #{} inside gettext strings

mzazrivec:
`#{}` inside a gettext string wont work, use `%{}` instead.

### DIFF
--- a/app/models/manageiq/providers/azure/cloud_manager/event_catcher.rb
+++ b/app/models/manageiq/providers/azure/cloud_manager/event_catcher.rb
@@ -7,7 +7,7 @@ class ManageIQ::Providers::Azure::CloudManager::EventCatcher < ManageIQ::Provide
   def self.all_valid_ems_in_zone
     super.select do |ems|
       ems.supports_timeline?.tap do |available|
-        _log.info(ems.unsupported_reason(:timeline) + _(" [#{ems.provider_region}]")) unless available
+        _log.info(ems.unsupported_reason(:timeline) + " [#{ems.provider_region}]") unless available
       end
     end
   end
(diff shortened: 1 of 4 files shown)


djberg96:
@mzazrivec Is this something you can add a spec for?

mzazrivec:
@djberg96 Probably yes, although this needs to be done as a shared spec for all repos, so the actual spec would have to land in core (i.e. it will be a different PR from this).

djberg96:
@mzazrivec Alright, once you create the shared spec (and it passes), then I'll merge.

mzazrivec:
> @mzazrivec Alright, once you create the shared spec (and it passes), then I'll merge.

To actually test what you're asking here, you need to extract the strings from the sources and parse the resulting catalog for `#{}`. So you're not going to see any passing spec here really, not unless somebody creates a PR with updated catalogs, which happens very rarely (the only person doing that is me).

mzazrivec:
> @mzazrivec Alright, once you create the shared spec (and it passes), then I'll merge.

Also, by definition of things, I cannot create a passing spec before this PR is merged.

djberg96:
@mzazrivec I see you added some changes. When it's ready to go let me know and I'll merge.

mzazrivec:
@djberg96 Yes, this one is good to go.

mzazrivec:
By the way, the spec we were discussing here got merged here: https://github.com/ManageIQ/manageiq/pull/17684 (which is why I had to include a new gettext catalog into this PR for it to pass CI).

djberg96:
Great, thanks!

djberg96:
@mzazrivec Note that this won't be backported without a BZ. If you need it backported, please create one. For now I've set the backport labels to "No".

mzazrivec:
Nah, I don't think a backport is needed here.

PR #274, opened 2018-07-03
Add support for custom endpoint

djberg96:
In order to support the Azure Stack environment we need to allow the users to set a custom endpoint since the default environments will not work. With the help of automatic discovery baked into the azure-armrest gem, we can support this fairly easily. If no endpoint is set then it simply defaults to finding the environment based on region as before.

Partially addresses https://bugzilla.redhat.com/show_bug.cgi?id=1594381

We will still need a modification on the UI side as well.

~~WIP for now until I add some specs.~~

### DIFF
--- a/app/models/manageiq/providers/azure/manager_mixin.rb
+++ b/app/models/manageiq/providers/azure/manager_mixin.rb
@@ -6,7 +6,8 @@ module ManageIQ::Providers::Azure::ManagerMixin
 
     client_id  = options[:user] || authentication_userid(options[:auth_type])
     client_key = options[:pass] || authentication_password(options[:auth_type])
-    self.class.raw_connect(client_id, client_key, azure_tenant_id, subscription, options[:proxy_uri] || http_proxy_uri, provider_region)
+
+    self.class.raw_connect(client_id, client_key, azure_tenant_id, subscription, options[:proxy_uri] || http_proxy_uri, provider_region, default_endpoint)
   end
 
   def verify_credentials(_auth_type = nil, options = {})
@@ -14,7 +15,7 @@ module ManageIQ::Providers::Azure::ManagerMixin
   end
 
   module ClassMethods
-    def raw_connect(client_id, client_key, azure_tenant_id, subscription, proxy_uri = nil, provider_region = nil)
+    def raw_connect(client_id, client_key, azure_tenant_id, subscription, proxy_uri = nil, provider_region = nil, endpoint = nil)
       require 'azure-armrest'
 
       if subscription.blank?
@@ -23,6 +24,18 @@ module ManageIQ::Providers::Azure::ManagerMixin
 
       if provider_region.blank?
         $azure_log.warn("No region selected. Validating credentials against public environment.")
+      end
+
+      endpoint_url = endpoint.respond_to?(:url) ? endpoint.url : endpoint.to_s
+
+      if endpoint_url.present?
+        begin
+          environment = ::Azure::Armrest::Environment.discover(:url => endpoint_url, :proxy => proxy_uri)
+        rescue SocketError
+          raise MiqException::MiqUnreachableError, _("Invalid endpoint")
+        end
+      else
+        environment = environment_for(provider_region)
       end
 
       ::Azure::Armrest::Configuration.log = $azure_log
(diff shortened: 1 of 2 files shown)


spec/models/manageiq/providers/azure/cloud_manager_spec.rb
```diff
@@ -76,6 +76,53 @@
...
+        @e.provider_region = "westus2"
+        @e.connect
+      end
+
+      it "with endpoint" do
+        expect(described_class).to receive(:raw_connect) do |clientid, clientkey, azure_tenant_id, subscription, http_proxy_uri, provider_region, endpoint|
+          expect(clientid).to eq("klmnopqrst")
+          expect(clientkey).to eq("1234567890")
+          expect(azure_tenant_id).to eq("abcdefghij")
+          expect(subscription).to eq("fghij67890")
```
bronaghs: @djberg96 why are lines 111  through 114  repeated so many times in this spec?

spec/models/manageiq/providers/azure/cloud_manager_spec.rb
```diff
@@ -76,6 +76,53 @@
...
+          expect(http_proxy_uri).to eq(nil)
+          expect(provider_region).to eq(nil)
+          expect(endpoint.url).to eq('http://www.foo.bar')
+          expect(endpoint.path).to eq('/')
+        end
+
+        @e.subscription = "fghij67890"
+        @e.endpoint = 'http://www.foo.bar'
+        @e.connect
+      end
```
bronaghs: @djberg96 - how about a test for a valid endpoint *with* a path?

spec/models/manageiq/providers/azure/cloud_manager_spec.rb
djberg96: @bronaghs Just to make sure they weren't negatively affected. I can of course remove them if you feel they're redundant.

spec/models/manageiq/providers/azure/cloud_manager_spec.rb
djberg96: Ok, will add that.

djberg96:
@bronaghs Ok, updated the specs, and even added a few more.

djberg96:
@bronaghs @Ladas I actually want to undo some things first. It turns out there's already a `default_endpoint` method that will work for us, so this PR is a bit overworked atm.

app/models/manageiq/providers/azure/manager_mixin.rb
```diff
@@ -25,6 +26,18 @@ def raw_connect(client_id, client_key, azure_tenant_id, subscription, proxy_uri
         $azure_log.warn("No region selected. Validating credentials against public environment.")
       end
 
+      endpoint_url = endpoint.respond_to?(:url) ? endpoint.url : endpoint.to_s
```
bronaghs: `endpoint.try(:url) || endpoint.to_s` - maybe a little simpler

app/models/manageiq/providers/azure/manager_mixin.rb
```diff
@@ -25,6 +26,18 @@ def raw_connect(client_id, client_key, azure_tenant_id, subscription, proxy_uri
...
+
+      if endpoint_url.present?
+        begin
+          environment = ::Azure::Armrest::Environment.discover(:url => endpoint_url, :proxy => proxy_uri)
+        rescue SocketError
+          raise MiqException::MiqInvalidCredentialsError, _("Invalid endpoint")
+        end
```
bronaghs: Is it always a credentials error?

djberg96: It basically means the endpoint is invalid, which I'm counting as "credentials" for our purposes. I actually couldn't find a better error, but I'm open to suggestions.

app/models/manageiq/providers/azure/manager_mixin.rb
djberg96: True, though I'm hoping https://github.com/ManageIQ/manageiq/pull/17678 will get merged so I can just call `to_s`.

app/models/manageiq/providers/azure/manager_mixin.rb
djberg96: Maybe this should be switched to an `MiqUnreachableError`.

bronaghs: Agreed.

app/models/manageiq/providers/azure/manager_mixin.rb
bronaghs: Ok, but if ManageIQ/manageiq#17678  doesn't end up getting merged, can you simplify this line?

PR #275, opened 2018-07-10
[WIP] Use valid api-version string if settings.yml value is invalid

djberg96:
If a user sets an invalid api-version string in the config/settings.yml file, then the refresh parser will bomb with an error from Azure. This could happen because they modified it manually, or because their environment has different values that are considered valid from what we specify by default.

This PR handles that possibility gracefully by checking the api-version specified in the settings.yml file versus the list of possible values that the Azure REST API tells us are valid. If there's a match, then things proceed normally.

If not, however, then we pull the most recent api-version string from the list of valid values and use that instead, logging a warning for good measure.

WIP until I can add some specs.


djberg96:
Note that because this makes one extra REST API call internally, the specs will fail using the current cassettes. Those will need to be regenerated, but atm we cannot because of current internal issues.

app/models/manageiq/providers/azure/refresh_helper_methods.rb
```diff
@@ -105,56 +105,78 @@ def network_routers
...
+    if valid_api_versions.include?(config_api_version)
+      config_api_version
+    else
+      valid_version_string = valid_api_versions.first
+
+      message = "Invalid api-version setting of '#{config_api_version}' for " \
+        "#{service.provider}/#{service.service_name} for EMS #{@ems.name}; " \
+        "using '#{valid_version_string}' instead."
+
+      _log.warn(message)
```
Fryguy: Is it possible to detect this at the time it's set in the settings?  I can show you where to do it...my only concern is if it makes a call out to the provider (can't tell from this code if it does or not).

djberg96: It does make one extra call internally to get that list.

bronaghs: I would like to see how to do that @Fryguy

djberg96:
@Ladas Not sure why, but another targeted refresh spec is failing after regenerating the cassettes, so I commented out the offending spec for now. Something about the vm from an orchestration template it doesn't like, but I'm not sure what.

Other than that, it's ready to go.

app/models/manageiq/providers/azure/refresh_helper_methods.rb
```diff
@@ -116,56 +116,83 @@ def network_routers
   end
 
   # Create the necessary service classes and lock down their api-version
-  # strings using the config/settings.yml from the provider repo. The
-  # "to_s" call for the version strings puts the date in the format
-  # that we need, i.e. "YYYY-MM-DD".
+  # strings using the config/settings.yml from the provider repo.
+
+  def cached_resource_provider_service(config)
+    @cached_resource_provider_service ||= resource_provider_service(config)
```
Ladas: if this is something that rarely changes, we could probably wrap it to cache_with_timeout block?

djberg96: I didn't know about `cache_with_timeout` but perhaps we can switch it to that if you prefer.

Ladas: yeah, for the things that do not change (or not frequently) it's perfect

PR #278, opened 2018-07-16
Default to StandardError if a connection cannot be made

djberg96:
Currently the `connection_rescue_block` is defaulting to a `MiqCredentialsError` as a last resort when an unexpected connection error occurs. As per @jrafanie's advice, this should be changed to re-raise `StandardError` so that more graceful handling will occur via the `AuthenticationMixin` module from the core repo, since it is now a "valid" error.

It's also just more sensical since by the time we fall into this bucket, since it's probably not an actual credentials error.

While we obviously cannot solve problems on Azure's end (which is typically when this error will be reached), we can handle it more gracefully with this change.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1600968

### DIFF
--- a/app/models/manageiq/providers/azure/manager_mixin.rb
+++ b/app/models/manageiq/providers/azure/manager_mixin.rb
@@ -50,7 +50,7 @@ module ManageIQ::Providers::Azure::ManagerMixin
       raise # Raise before falling into catch-all block below
     rescue StandardError => err
       _log.error("Error Class=#{err.class.name}, Message=#{err.message}, Backtrace=#{err.backtrace}")
-      raise MiqException::MiqInvalidCredentialsError, _("Unexpected response returned from system: %{error_message}") % {:error_message => err.message}
+      raise err, _("Unexpected response returned from system: %{error_message}") % {:error_message => err.message}
     end
 
     def environment_for(region)
--- a/spec/models/manageiq/providers/azure/cloud_manager_spec.rb
+++ b/spec/models/manageiq/providers/azure/cloud_manager_spec.rb
@@ -102,7 +102,7 @@ describe ManageIQ::Providers::Azure::CloudManager do
       end
       it "handles unknown error" do
         allow(Azure::Armrest::Configuration).to receive(:new).and_raise(StandardError)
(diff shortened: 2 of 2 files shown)


djberg96:
@jrafanie @bronaghs @01100010011001010110010101110000 ^

app/models/manageiq/providers/azure/manager_mixin.rb
```diff
@@ -50,7 +50,7 @@ def connection_rescue_block
       raise # Raise before falling into catch-all block below
     rescue StandardError => err
       _log.error("Error Class=#{err.class.name}, Message=#{err.message}, Backtrace=#{err.backtrace}")
-      raise MiqException::MiqInvalidCredentialsError, _("Unexpected response returned from system: %{error_message}") % {:error_message => err.message}
+      raise MiqException::MiqUnreachableError, _("Unexpected response returned from system: %{error_message}") % {:error_message => err.message}
```
jrafanie: Perhaps it makes sense to just re-raise err because we're not really sure it's unreachable at this point.  We treat `StandardError` exceptions the same as `MiqUnreachableError`: we retry more frequently with an exponential backoff.  The problem was that we were explicitly raising `MiqInvalidCredentialsError`, an exception we don't retry as frequently.

https://github.com/ManageIQ/manageiq/blob/d101c7997d556f08a7b0747703654ba74915b7da/app/models/mixins/authentication_mixin.rb#L401-L409

djberg96: @jrafanie Fair enough, I'll change it.

simaishi:
@djberg96 Can this be `gaprindashvili/yes`?

djberg96:
@simaishi Yes, tagged.

simaishi:

```
$ git log -1
commit 78cb5e84e41f9969e37eca7283a7653b62636562
Date:   Mon Jul 16 14:33:34 2018 -0400

    Merge pull request #278 from djberg96/unreachable
    
    Default to StandardError if a connection cannot be made
    (cherry picked from commit 745543cdfd81c3883a93f2f1cc59caaa86e31ad6)
    
    Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1601589
```

PR #280, opened 2018-07-19
Enable graph refresh by default

Ladas:
Enable graph refresh by default


djberg96:
@Ladas I'm all for it, but I think it's above my pay grade. ;)

BTW, I'm assuming this should NOT be backported, correct?

Ladas:
@bronaghs no backport right? We will enable it in G- release only when there are problems, right?

bronaghs:
@Ladas - right.

PR #281, opened 2018-07-23
InventoryCollection builder_params -> default_values

slemrmartin:
**Issue**: https://github.com/ManageIQ/manageiq/issues/17396

- [x] **depends on** https://github.com/ManageIQ/manageiq/pull/17742

Removing "builder_params" term from all inventory_collection related classes


slemrmartin:
cc @agrare

PR #291, opened 2018-09-20
Fix parent association in graph refresh

djberg96:
The when setting a parent, the `parent_ems_ref` helper method pulls its information straight from the JSON properties. Consequently, we need to explicitly downcase it to ensure that the association is set correctly.

### DIFF
--- a/app/models/manageiq/providers/azure/inventory/parser/cloud_manager.rb
+++ b/app/models/manageiq/providers/azure/inventory/parser/cloud_manager.rb
@@ -72,7 +72,7 @@ class ManageIQ::Providers::Azure::Inventory::Parser::CloudManager < ManageIQ::Pr
       series = persister.flavors.find(instance.properties.hardware_profile.vm_size.downcase)
 
       rg_ems_ref = collector.get_resource_group_ems_ref(instance)
-      parent_ref = collector.parent_ems_ref(instance).try(:downcase)
+      parent_ref = collector.parent_ems_ref(instance)
 
       # We want to archive VMs with no status
       next if (status = collector.power_status(instance)).blank?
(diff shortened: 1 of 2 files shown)


app/models/manageiq/providers/azure/inventory/parser/cloud_manager.rb
```diff
@@ -72,7 +72,7 @@ def instances
       series = persister.flavors.find(instance.properties.hardware_profile.vm_size.downcase)
 
       rg_ems_ref = collector.get_resource_group_ems_ref(instance)
-      parent_ref = collector.parent_ems_ref(instance)
+      parent_ref = collector.parent_ems_ref(instance).try(:downcase)
```
agrare: @djberg96 are all possible parent ems_refs downcased?  Could we just do this in `parent_ems_ref`?  Want to avoid someone else using this method without knowing they need to downcase the result to be able to reference something.

djberg96: Unfortunately, we only did it for the managed images, not the unmanaged ones. I could modify it there, but then I would need to modify the parser again, plus add a data migration.

agrare: No that's fine, how about [here](https://github.com/ManageIQ/manageiq-providers-azure/blob/20b52b9d9b320005b035bff7c21a7c2aae54c137/app/models/manageiq/providers/azure/refresh_helper_methods.rb#L49) ?  Could we do `.try(:image_reference).try(:id).try(:downcase)`

djberg96: Yeah, that should work. I'll update it.

PR #295, opened 2018-10-04
Rebuild spec environment and update specs

djberg96:
Because we no longer have access to our original dev environment, I've had to rebuild resources within Azure for testing. The original environment for testing was added over several years with no real structure.

This PR adds a bash script that can be used to rebuild the entire environment from scratch. Using the updated environment, I've also updated the specs.

This is a WIP for now until I figure out how to regenerate some of the cassettes, as well as deal with the "can't quote Hash" errors that I'm seeing.


djberg96:
@Ladas @tumido This is about as far as I can get. At the moment the specs are passing, but you'll notice that I've had to comment out a few, namely a few of the load balancer + vm specs, as well as the calls to `assert_models_not_changed`.

Any and all help appreciated. If you feel these aren't crucial for now, I would like to proceed so that I can get to the endpoint PR's. If they are crucial, it may just be a matter of updating the shell scripts and/or templates used to create these resources.

spec/spec_helper.rb
```diff
@@ -12,6 +12,7 @@
       VCR.request_matchers.uri_without_param('api-version')
     ]
   }
+  #config.debug_logger = File.open('vcr.debug', 'w')
```
Ladas: is this on purpose?

Ladas:
@djberg96 lets just ignore specific tables in assert_models_not_changed and put there TODO and I am good with merging this.

Ladas:
@djberg96 looks like there is still some valid CI failure with labels

spec/spec_helper.rb
djberg96: Yes, otherwise I always have to look it up when I need it. :)

djberg96:
@Ladas Ah, I see what happened with the label failures. I'll update that.

Ladas:
@djberg96 ok, lets put this PR to passing state and we can fix the rest of the specs in a followup

djberg96:
@Ladas It's passing for me locally, both on Linux and Mac, so I'm not sure what the current failures are about yet. Investigating...

Ladas:
@djberg96 hm, before the local vs. Ci failures were caused by obfuscating the data. So locally I'de see it if I remove the secrets file content

djberg96:
@Ladas Alright, had to comment out the template specs for some reason, not sure why they fail atm. Anyway, it's green now. :)

Ladas:
@djberg96 I think that we are just disabling the cfme check for VCRs in other providers, so we should do it here too.

PR #298, opened 2018-10-26
[WIP] Only select the VM fields that Azure uses

djberg96:
This is a minor performance enhancement that I worked on after realizing that a significant number of columns in the `vms` table aren't used by Azure. Many only apply to VMWare. So I added a `default_scope` on the `ManageIQ::Providers::Azure::CloudManager::Vm` model.

The net result is a 16 byte reduction per Vm object (128 vs 144), as reported by `ObjectSpace.memsize_of`, in my local testing. While individually small, this would add up with significant numbers of Vm's. It also makes inspection easier in the console, since it only shows the fields we care about.

This is a WIP for now, as it appears to blow up a custom `calculate` method that was added as a core extension:

https://github.com/ManageIQ/manageiq/blob/master/lib/extensions/ar_virtual.rb#L781

### DIFF
--- a/app/models/manageiq/providers/azure/cloud_manager/vm.rb
+++ b/app/models/manageiq/providers/azure/cloud_manager/vm.rb
@@ -15,6 +15,7 @@ class ManageIQ::Providers::Azure::CloudManager::Vm < ManageIQ::Providers::CloudM
     cpu_shares_level
     deprecated
     fault_tolerance
+    format
     linked_clone
     memory_hot_add_enabled
     memory_hot_add_increment
@@ -24,6 +25,7 @@ class ManageIQ::Providers::Azure::CloudManager::Vm < ManageIQ::Providers::CloudM
     memory_reserve_expand
     memory_shares
     memory_shares_level
+    registered
     tools_status
   ].freeze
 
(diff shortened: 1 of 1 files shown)


app/models/manageiq/providers/azure/cloud_manager/vm.rb
```diff
@@ -2,6 +2,32 @@ class ManageIQ::Providers::Azure::CloudManager::Vm < ManageIQ::Providers::CloudM
...
+      :publicly_available,
+      :raw_power_state,
+      :resource_group_id,
+      :state_changed_on,
+      :tenant_id,
+      :type,
+      :uid_ems,
+      :vendor
+    ])
+  end
```
Fryguy: This seems like a really bad idea.  If we add a core column we have to remember to add it here.  What kind of *actual* performance gain / memory improvments are we talking about?

cc @bdunne @agrare

djberg96:
@Fryguy should https://github.com/ManageIQ/manageiq/blob/master/lib/extensions/ar_virtual.rb#L781 be modified to explicitly include the arguments?

djberg96:
Closing for now since I think the very act of scoping countered any memory savings I was seeing. That and I'm not doing provider work any more.

PR #300, opened 2018-10-29
Handle possibility that disk might not have sku

djberg96:
Apparently it's possible for a managed disk to not have a sku. This handles that possibility.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1642478

### DIFF
--- a/app/models/manageiq/providers/azure/cloud_manager/refresh_parser.rb
+++ b/app/models/manageiq/providers/azure/cloud_manager/refresh_parser.rb
@@ -399,7 +399,7 @@ module ManageIQ::Providers
             blob_name = uri.basename
 
             storage_acct = @storage_accounts.find { |s| s.name.casecmp(storage_name).zero? }
-            mode = storage_acct.sku.name
+            mode = storage_acct.try(:sku).try(:name)
 
             if @options.get_unmanaged_disk_space && disk_size.nil?
               storage_keys = @sas.list_account_keys(storage_acct.name, storage_acct.resource_group)
--- a/app/models/manageiq/providers/azure/cloud_manager/refresh_parser.rb
+++ b/app/models/manageiq/providers/azure/cloud_manager/refresh_parser.rb
@@ -381,7 +381,7 @@ module ManageIQ::Providers
 
           if managed_disk
             disk_size = managed_disk.properties.disk_size_gb.gigabytes
-            mode      = managed_disk.sku.name
+            mode      = managed_disk.try(:sku).try(:name)
           else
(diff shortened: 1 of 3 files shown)


bronaghs:
what about unmanaged disks? Maybe add a spec too

djberg96:
@bronaghs We don't try to access the sku for unmanaged disks (don't think it has one).

djberg96:
@bronaghs Upon further review, yes we the same thing for unmanaged. This was already in place for graph refresh, so I've added it for classic refresh as well.

@Ladas Not sure why specs are failing on Travis. They pass for me locally, even after removing my secrets.yml file.

djberg96:
@bronaghs @Ladas Ok, commented out the test failing on targeted refresh specs for now, will come back to it later.

In any case, this fix was tested in the customer's dev environment and solved their issue.

djberg96:
@Ladas Not sure, when I look at the disks table for a local instance, they all seem to have a `mode` set, except one. That one now has two entries - one with a mode and one without. And they seem to be backing up the same VM, which has two entries with identical `ems_ref`'s.

djberg96:
@Ladas Ok, further debugging has revealed that there's an issue with unmanaged storage accounts and targeted refresh. The short version is that it's using the attached VM's resource group instead of its own resource group for lookup. If those happen to be different, then the lookup fails and it sets the mode to nil.

Since it's a separate issue from this, so I'm asking you to merge that and I'll create a different github issue.

simaishi:
Hammer backport details:

```
$ git log -1
commit 07691319edf14af6f6edf64c382f37fd616bc072
Author: Ladislav Smola <lsmola@redhat.com>
Date:   Thu Nov 1 17:02:53 2018 +0100

    Merge pull request #300 from djberg96/disk_sku
    
    Handle possibility that disk might not have sku
    
    (cherry picked from commit 60d75641b09d7c79e8c27890ca39c772a9836b94)
    
    Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1642478
```

simaishi:

```
$ git log -1
commit 214e1a51cbea600d7036270069cba878d05553d2
Author: Ladislav Smola <lsmola@redhat.com>
Date:   Thu Nov 1 17:02:53 2018 +0100

    Merge pull request #300 from djberg96/disk_sku
    
    Handle possibility that disk might not have sku
    
    (cherry picked from commit 60d75641b09d7c79e8c27890ca39c772a9836b94)
    
    Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1646646
```

PR #302, opened 2018-11-05
Handle regions where metrics are unsupported

djberg96:
It seems that even if a region has the Microsoft.Insights provider enabled, it still does not necessarily support collection of certain metrics in certain regions, e.g. Germany.

In practice what we see is a `Azure::Armrest::BadRequestException` raised in such cases, combined with a misleading error message from Azure. The error message suggests that it's simply a matter of adjusting the `api-version` string, but in reality none of them work if the region simply isn't supported.

Instead of raising an error, this PR adjusts the `MetricsCapture#get_counters` method to explicitly check for a `Azure::Armrest::BadRequestException`, logs a warning, and returns an empty array.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1645117

### DIFF
--- a/app/models/manageiq/providers/azure/cloud_manager/metrics_capture.rb
+++ b/app/models/manageiq/providers/azure/cloud_manager/metrics_capture.rb
@@ -209,6 +209,11 @@ class ManageIQ::Providers::Azure::CloudManager::MetricsCapture < ManageIQ::Provi
           .list('Microsoft.Compute', 'virtualMachines', target.name, target.resource_group.name)
           .select { |m| m.name.value.in?(COUNTER_NAMES) }
       end
+    rescue ::Azure::Armrest::BadRequestException # Probably means region is not supported
+      msg = "Problem collecting metrics for #{target.name}/#{target.resource_group.name}. "\
+            "Region [#{ems.provider_region}] may not be supported."
+      _log.warn(msg)
+      counters = []
     rescue ::Azure::Armrest::RequestTimeoutException # Problem on Azure side
       _log.warn("Timeout attempting to collect metrics definitions for: #{target.name}/#{target.resource_group.name}. Skipping.")
       counters = []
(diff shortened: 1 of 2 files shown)


JPrause:
@miq-bot add_label blocker

JPrause:
@bronaghs as this is for a blocker, can you review asap.

app/models/manageiq/providers/azure/cloud_manager/metrics_capture.rb
```diff
@@ -209,6 +209,11 @@ def get_counters(metrics_conn)
           .list('Microsoft.Compute', 'virtualMachines', target.name, target.resource_group.name)
           .select { |m| m.name.value.in?(COUNTER_NAMES) }
       end
+    rescue ::Azure::Armrest::BadRequestException # Probably means region is not supported
```
agrare: Is there anything more specific on why it failed in the message that we could check?  something like `raise unless err.message =~ /Region does not support metrics/` or whatever the exception message says.

agrare: > The error message suggests that it's simply a matter of adjusting the api-version string, but in reality none of them work if the region simply isn't supported.

Oh hm okay

djberg96: @agrare yeah, unfortunately the error message is not friendly.

djberg96:
@agrare More difficult than you would think, since it may just be a matter of the Microsoft.Insights service not being enabled (which we already check for). What could happen is that the Insights service isn't enabled, the region would be tagged as "unsupported", then later they enable Insights and wonder why metrics aren't being collected.

simaishi:
Hammer backport details:

```
$ git log -1
commit 3e5ba61abfd42d3a3b13d135b2aa2a04c98b4162
Date:   Thu Nov 8 10:55:32 2018 -0500

    Merge pull request #302 from djberg96/metrics_unsupported
    
    Handle regions where metrics are unsupported
    
    (cherry picked from commit a94f2b03a21ac70f07e11ef327740e6af8d889ce)
    
    Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1645117
```

simaishi:

```
$ git log -1
commit 65952ca7b547d1f9bb9f74fc4d52124ab700963b
Date:   Thu Nov 8 10:55:32 2018 -0500

    Merge pull request #302 from djberg96/metrics_unsupported
    
    Handle regions where metrics are unsupported
    
    (cherry picked from commit a94f2b03a21ac70f07e11ef327740e6af8d889ce)
    
    Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1648955
```

PR #307, opened 2018-11-16
Always set vm_or_template.connection_state

agrare:
Instead of defaulting to nil always set vm_or_template.connection_state


djberg96:
@agrare I'm fine with this, though I'm curious what motivated it.

agrare:
@djberg96 there was some confusion around the handling of connection_state.nil? where it was showing cloud vms as disconnected.  Turns out every cloud provider override the disconnected method instead of having the parser set "connected".

djberg96:
@agrare Ok, thanks.

PR #310, opened 2018-12-04
Don't stub all_ems_in_zone

kbrock:
Depends on:
- [x] ManageIQ/manageiq#18266

I want to remove the `to_a` from `Ems#all_ems_in_zone`
So I'm removing the stub in the specs so the real `all_ems_in_zone` is called, tested and able to return a relation.

@miq-bot add_label tests

### DIFF
--- a/spec/models/manageiq/providers/azure/cloud_manager/event_catcher_spec.rb
+++ b/spec/models/manageiq/providers/azure/cloud_manager/event_catcher_spec.rb
@@ -3,8 +3,8 @@ describe ManageIQ::Providers::Azure::CloudManager::EventCatcher do
     let(:unsupported_reason) { "Timeline events not supported for this region" }
 
     before do
-      @ems = FactoryBot.create(:ems_azure)
-      allow(@ems).to receive(:authentication_status_ok?).and_return(true)
+      server = EvmSpecHelper.local_miq_server
+      @ems = FactoryGirl.create(:ems_azure, :with_authentication, :zone => server.zone)
       allow(described_class).to receive(:all_ems_in_zone).and_return([@ems])
     end
 
--- a/spec/models/manageiq/providers/azure/cloud_manager/event_catcher_spec.rb
+++ b/spec/models/manageiq/providers/azure/cloud_manager/event_catcher_spec.rb
@@ -5,18 +5,17 @@ describe ManageIQ::Providers::Azure::CloudManager::EventCatcher do
     before do
       server = EvmSpecHelper.local_miq_server
       @ems = FactoryGirl.create(:ems_azure, :with_authentication, :zone => server.zone)
-      allow(described_class).to receive(:all_ems_in_zone).and_return([@ems])
     end
 
     it "returns a valid ems for zone if timeline events are supported" do
-      allow(@ems).to receive(:supports_timeline?).and_return(true)
+      expect_any_instance_of(@ems.class).to receive(:supports_timeline?).and_return(true)
       expect($log).not_to receive(:info).with(/#{unsupported_reason}/)
       expect(described_class.all_valid_ems_in_zone).to include(@ems)
     end
 
     it "returns an empty list for zone if timeline events are not supported" do
-      allow(@ems).to receive(:supports_timeline?).and_return(false)
-      allow(@ems).to receive(:unsupported_reason).and_return(unsupported_reason)
+      expect_any_instance_of(@ems.class).to receive(:supports_timeline?).and_return(false)
+      expect_any_instance_of(@ems.class).to receive(:unsupported_reason).and_return(unsupported_reason)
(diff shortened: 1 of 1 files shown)


djberg96:
@kbrock rebase?

spec/models/manageiq/providers/azure/cloud_manager/event_catcher_spec.rb
```diff
@@ -3,20 +3,19 @@
     let(:unsupported_reason) { "Timeline events not supported for this region" }
 
     before do
-      @ems = FactoryBot.create(:ems_azure)
-      allow(@ems).to receive(:authentication_status_ok?).and_return(true)
-      allow(described_class).to receive(:all_ems_in_zone).and_return([@ems])
```
kbrock: my goal is to remove this stub and instead use the `authentication_status_ok?` from the database.

spec/models/manageiq/providers/azure/cloud_manager/event_catcher_spec.rb
```diff
@@ -3,20 +3,19 @@
...
+      server = EvmSpecHelper.local_miq_server
+      @ems = FactoryGirl.create(:ems_azure, :with_authentication, :zone => server.zone)
     end
 
-    it "returns a valid ems for zone if timeline events are supported" do
-      allow(@ems).to receive(:supports_timeline?).and_return(true)
+    it "returns a valid ems for zone when timeline events are supported" do
+      expect_any_instance_of(@ems.class).to receive(:supports_timeline?).and_return(true)
```
kbrock: a. I don't know if these tests are even valid.
I see them in other places, but since this method was not defined here, seems moot.
b. Is it possible for an azure ems to not support timelines?

c. rubocop does not like this. Not sure that the complain should stop us from merging but...

The alternative is to not stub it at all. Does anyone know how to setup an ems so it either does or does not `supports_timeline?` ? (again, is this even valid?)

kbrock:
unwip: Not optimal, but this removes stubbing of the method I want to change

djberg96:
@kbrock At the moment it looks like `supports_timeline?` is always true since it's inherited from core. We *could* check it for Azure by validating that the Microsoft.Insights provider has been registered first, but that's not what we do now. Not at the EMS level, anyway.

kbrock:
@djberg96 thanks - THAT is why I couldn't do that.

So this test was generic - and pointless.

I can remove the first spec

kbrock:
> At the moment it looks like supports_timeline? is always true

Is there something special I need to do.
When I remove that stub to the `allow` it returns false.

kbrock:
@miq-bot add_label test

Is there anything holding this back?
(I'm not actively trying to refactor the base of this test, but I'm sure I'll revisit within the next year)

spec/models/manageiq/providers/azure/cloud_manager/event_catcher_spec.rb
djberg96: Regarding `b)`, it's possible if certain services are disabled on the subscription. I don't see that ever happening in practice though. Stack is another matter, but that's a separate provider now.

kbrock:
Ok, wanted to improve `all_valid_ems_in_zone` so that instead of using a `select {}`, it uses a `where()` and `select {}` or optimally just a `where`

but in order for to use a `where()`, `all_ems_in_zone` needs to return a scope.

We can do this by making our test data work so the values returned from the database in `all_ems_in_zone` have properly set `authentication_status_ok?`, or we stub all calls into `authentication_status_ok?` to have it return what we want.

I was not able to create the EMS records correctly. (Which I found frustrating)

I'm also a little confused why this method is even tested in this provider. It is possible that this method was defined in the providers and moved up?

djberg96:
Yeah, I've been fighting with it as well. As for why it's tested, I imagine it's because the EventCatcher in this provider has defined its own `all_valid_ems_in_zone` method.

kbrock:
Ahh, amazon provider overrides `all_valid_ems_in_zone` which is a duplicate for the parent. (probably what I was thinking)

djberg96:
@kbrock Well, let's take a step back. As far as I can tell, removing `to_a` at https://github.com/ManageIQ/manageiq/blob/master/app/models/mixins/per_ems_worker_mixin.rb#L23 doesn't break this spec.

So, do we really need to change it? Or did I miss something?

If you want FactoryBot to return a relation, we can use `FactoryBot.create_list` instead.

djberg96:
As per our gitter conversation, I think the best thing to do is to just remove the `to_a` from core, confirm that it doesn't cause any breakage using the cross-repo test framework, and close this once it's been confirmed that didn't cause any other issues.

kbrock:
sigh. alas. not worth it to fight the 👮 (rubocop)

PR #321, opened 2019-01-09
[WIP] Check for service support

djberg96:
Previously we were checking certain services for support, and then only in classic refresh. This PR takes a more generic approach for targeted refresh by modifying the `gather_data_for_this_region` method so that each service name is checked before attempting to collect data.

To keep the number of external REST API calls to a minimum, we use the cached resource provider service, and cache the results of each service check since it is extremely unlikely that the status of a service will change once it's set for a given instance. For such rare situations, the instance could be restarted.

WIP for now as I'm not sure if the cassettes will need to be regenerated.

https://bugzilla.redhat.com/show_bug.cgi?id=1566600


Fryguy:
Travis status is stuck, so I'm bouncing this PR.

djberg96:
@Fryguy thanks, but I'll have to regenerate the cassettes for this one either way unfortunately.

djberg96:
Closing, since this really only occurred in azure-stack, and since stack is now a separate provider (which handles this).

PR #325, opened 2019-03-19
Remove references to MiqPassword

Fryguy:
@agrare @djberg96 Please review.  This gets us closer to removing MiqPassword from manageiq-gems-pending.

cc @jrafanie


agrare:
@Fryguy does this depend on https://github.com/ManageIQ/manageiq/pull/18566 or does gems-pending also define `ManageIQ::Password` ?

PR #328, opened 2019-04-04
Use string class name instead of hitting autoload

jrafanie:
Fixes a deprecation on rails 5.1:
DEPRECATION WARNING: Passing a class to the `class_name` is deprecated and will raise an ArgumentError in Rails 5.2. It eagerloads more classes than necessary and potentially creates circular dependencies. Please pass the class name as a string: `has_many :floating_ips, class_name: 'ManageIQ::Providers::Azure::NetworkManager::FloatingIp'` (called from <class:NetworkManager> at /Users/joerafaniello/.gem/ruby/2.4.5/bundler/gems/manageiq-providers-azure-7d83a11a9ca9/app/models/manageiq/providers/azure/network_manager.rb:20)


jrafanie:
Related to https://github.com/ManageIQ/manageiq/pull/18076

djberg96:
👍

jrafanie:
@djberg96 green merge button?  🍰 🍪

PR #336, opened 2019-06-28
Use ems leafs in specs 

d-m-u:
Found here: https://github.com/ManageIQ/manageiq-providers-amazon/pull/542

One of these specs is pointless now since the leaf will always create two. 

Broken originally by https://github.com/ManageIQ/manageiq/pull/18842


d-m-u:
@miq-bot assign @agrare

d-m-u:
@miq-bot add_label test

PR #337, opened 2019-07-02
Add a rubocop channel to the codeclimate config

djberg96:
See ManageIQ/manageiq#18840 for more information

Unfortunately there's no way to use a centralized .codeclimate, so we have to configure the configuration files individually.


djberg96:
@miq-bot add_reviewer @agrare

PR #341, opened 2019-07-30
capture memory metric

AlexanderZagaynov:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1724415

### DIFF
--- a/app/models/manageiq/providers/azure/cloud_manager/metrics_capture.rb
+++ b/app/models/manageiq/providers/azure/cloud_manager/metrics_capture.rb
@@ -86,18 +86,15 @@ class ManageIQ::Providers::Azure::CloudManager::MetricsCapture < ManageIQ::Provi
     Benchmark.realtime_block(:connect) {}
 
     ems.with_provider_connection do |connection|
-      # TODO: code smell - should be in a 'collector' class instantiated here / find a better solution?
-      @connection = connection
-      @start_time = start_time.iso8601
-      @end_time   = end_time.iso8601
+      base_url = connection.environment.resource_url
 
       ## Counters from the metrics api
 
-      metrics_conn = Azure::Armrest::Insights::MetricsService.new(@connection)
+      metrics_conn = Azure::Armrest::Insights::MetricsService.new(connection)
 
       metrics, _timings = Benchmark.realtime_block(:capture_counters) do
         # azure-armrest gem needs to be modified to accept query ('list_metrics' method)
-        response = metrics_conn.send(:rest_get, metrics_url.to_s)
+        response = metrics_conn.send(:rest_get, metrics_url(base_url, start_time, end_time).to_s)
         Azure::Armrest::ArmrestCollection.create_from_response(response, Azure::Armrest::Insights::Metric)
       end
 
@@ -115,12 +112,12 @@ class ManageIQ::Providers::Azure::CloudManager::MetricsCapture < ManageIQ::Provi
 
       ## Counters, which are available only through legacy API and storage account (raw)
 
-      storage_conn = Azure::Armrest::StorageAccountService.new(@connection)
+      storage_conn = Azure::Armrest::StorageAccountService.new(connection)
       storage_accounts = storage_conn.list_all
 
       metric_definitions, _timings = Benchmark.realtime_block(:capture_counters) do
         # azure-armrest gem needs to be modified to accept version ('list_definitions' method)
-        response = metrics_conn.send(:rest_get, definitions_url.to_s)
+        response = metrics_conn.send(:rest_get, definitions_url(base_url).to_s)
         Azure::Armrest::ArmrestCollection.create_from_response(response, Azure::Armrest::Insights::MetricDefinition)
       end
 
@@ -140,17 +137,17 @@ class ManageIQ::Providers::Azure::CloudManager::MetricsCapture < ManageIQ::Provi
         storage_account_keys = storage_conn.list_account_keys(storage_account.name, storage_account.resource_group)
         storage_account_key  = storage_account_keys.fetch('key1')
 
+        filter = <<~FILTER
+          CounterName eq '#{metric_name}' AND \
+          PartitionKey eq '#{metric_location.partition_key}' AND \
+          Timestamp ge datetime'#{start_time.iso8601}' AND \
+          Timestamp le datetime'#{end_time.iso8601}'
+        FILTER
+
         metric_location.table_info.each do |table_info|
           t_start_time = parse_timestamp(table_info.start_time)
           t_end_time   = parse_timestamp(table_info.end_time).end_of_day
           next unless time_interval.overlaps?(t_start_time..t_end_time)
-
-          filter = [
-            "CounterName eq '#{metric_name}'",
-            "PartitionKey eq '#{metric_location.partition_key}'",
-            "Timestamp ge datetime'#{@start_time}'",
-            "Timestamp le datetime'#{@end_time}'",
-          ].join(' and ')
 
           table_data, _timings = Benchmark.realtime_block(:capture_counters) do
             storage_account.table_data(
@@ -176,9 +173,10 @@ class ManageIQ::Providers::Azure::CloudManager::MetricsCapture < ManageIQ::Provi
         next if metric_values.empty?
 
         # { timestamp => [value, ...], ... }
-        timestamped_values = metric_values.each_with_object({}) do |ts_values, memo|
+        timestamped_values = Hash.new { |h, k| h[k] = [] }
+        metric_values.each do |ts_values|
           ts_values.each do |timestamp, values|
-            memo[timestamp] = (memo[timestamp] || []) + values
+            timestamped_values[timestamp].concat(values)
           end
         end
 
@@ -257,21 +255,19 @@ class ManageIQ::Providers::Azure::CloudManager::MetricsCapture < ManageIQ::Provi
     ].join('/')
   end
 
-  def resource_url(tail_path = nil, query = {})
-    @resource_url ||= URI.join(@connection.environment.resource_url, resource_uri).freeze
-
-    url       = @resource_url.dup
+  def resource_url(base_url, tail_path = nil, query = {})
+    url       = URI.join(base_url, resource_uri)
     url.path  = [url.path, tail_path].join('/') if tail_path
     url.query = URI.encode_www_form(query) unless query.empty?
-
     url
   end
 
-  def metrics_url
-    @metrics_url ||= resource_url(
-      # https://docs.microsoft.com/en-us/rest/api/monitor/metrics/list#uri-parameters
+  def metrics_url(base_url, start_time, end_time)
+    resource_url(
+      base_url,
       'providers/microsoft.insights/metrics',
-      'timespan'    => "#{@start_time}/#{@end_time}",
+      # https://docs.microsoft.com/en-us/rest/api/monitor/metrics/list#uri-parameters
+      'timespan'    => "#{start_time.iso8601}/#{end_time.iso8601}",
       'interval'    => INTERVAL_1_MINUTE,
       'metricnames' => METRIC_NAMES[:api].join(','),
       'aggregation' => 'Average',
@@ -279,8 +275,9 @@ class ManageIQ::Providers::Azure::CloudManager::MetricsCapture < ManageIQ::Provi
     ).freeze
   end
 
-  def definitions_url
-    @definitions_url ||= resource_url(
+  def definitions_url(base_url)
+    resource_url(
+      base_url,
       'providers/microsoft.insights/metricDefinitions',
       'api-version' => '2015-07-01'
     ).freeze
--- a/app/models/manageiq/providers/azure/cloud_manager/metrics_capture.rb
+++ b/app/models/manageiq/providers/azure/cloud_manager/metrics_capture.rb
@@ -77,76 +77,135 @@ class ManageIQ::Providers::Azure::CloudManager::MetricsCapture < ManageIQ::Provi
       return
     end
 
-    end_time   = end_time   ? end_time.utc   : Time.now.utc
-    start_time = start_time ? start_time.utc : (end_time - 4.hours) # 4 hours for symmetry with VIM
+    end_time       = end_time   ? end_time.to_time.utc   : Time.now.utc
+    start_time     = start_time ? start_time.to_time.utc : (end_time - 4.hours) # 4 hours for symmetry with VIM
+    time_interval  = start_time..end_time
+    counter_values = {}
 
     # This is just for consistency, to produce a :connect benchmark
     Benchmark.realtime_block(:connect) {}
 
     ems.with_provider_connection do |connection|
-      metrics_conn = Azure::Armrest::Insights::MetricsService.new(connection)
+      # TODO: code smell - should be in a 'collector' class instantiated here / find a better solution?
+      @connection = connection
+      @start_time = start_time.iso8601
+      @end_time   = end_time.iso8601
 
       ## Counters from the metrics api
 
-      metrics_uri = URI.join(
-        metrics_conn.configuration.environment.resource_url,
-        "#{resource_uri}/providers/microsoft.insights/metrics"
-      )
-
-      # https://docs.microsoft.com/en-us/rest/api/monitor/metrics/list#uri-parameters
-      metrics_uri.query = URI.encode_www_form(
-        'timespan'    => "#{start_time.iso8601}/#{end_time.iso8601}",
-        'interval'    => INTERVAL_1_MINUTE,
-        'metricnames' => api_metric_names.join(','),
-        'aggregation' => 'Average',
-        'api-version' => metrics_conn.api_version
-      )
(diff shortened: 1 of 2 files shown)


AlexanderZagaynov:
@djberg96 @agrare please review.

P.S. I see this PR as a temporary step on a way to the Azure-native gems and total metrics refactoring (https://bugzilla.redhat.com/show_bug.cgi?id=1677564). That's why you can see a lot of workarounds in the specs - I don't want to waste much time there, considering that they will be thrown away with mentioned RFE.

spec/models/manageiq/providers/azure/cloud_manager/metrics_capture_spec.rb
```diff
@@ -4,58 +4,71 @@
   let(:ems)      { FactoryBot.create(:ems_azure) }
   let(:vm)       { FactoryBot.build(:vm_azure, :ext_management_system => ems, :ems_ref => "my_ems_ref") }
   let(:group)    { FactoryBot.build(:azure_resource_group) }
+  let(:armrest_environment) { double('Armrest environment', :resource_url => 'http://example.com') }
```
djberg96: I'd like people to use `instance_double` over `double` so that we can be sure our mocks are valid.

app/models/manageiq/providers/azure/cloud_manager/metrics_capture.rb
```diff
@@ -206,19 +257,36 @@ def resource_uri
...
+    @definitions_url ||= resource_url(
+      'providers/microsoft.insights/metricDefinitions',
+      'api-version' => '2015-07-01'
+    ).freeze
   end
 
-  def raw_metric_names
-    METRIC_NAMES[:raw]
+  def parse_timestamp(timestamp)
+    Time.parse(timestamp).utc
```
djberg96: Hardly seems worth being a separate method. Is it too pervasive to drop?

app/models/manageiq/providers/azure/cloud_manager/metrics_capture.rb
```diff
@@ -206,19 +257,36 @@ def resource_uri
...
+      'api-version' => '2018-01-01'
+    ).freeze
   end
 
-  def raw_counters_info
-    COUNTERS_INFO[:raw]
+  def definitions_url
+    @definitions_url ||= resource_url(
+      'providers/microsoft.insights/metricDefinitions',
+      'api-version' => '2015-07-01'
```
djberg96: Do we want to hard code this? I can't remember, but I thought we set in this in config/settings.yml.

Fryguy:
@agrare Please review

app/models/manageiq/providers/azure/cloud_manager/metrics_capture.rb
```diff
@@ -77,76 +77,135 @@ def perf_collect_metrics(interval_name, start_time = nil, end_time = nil)
...
+    time_interval  = start_time..end_time
+    counter_values = {}
 
     # This is just for consistency, to produce a :connect benchmark
     Benchmark.realtime_block(:connect) {}
 
     ems.with_provider_connection do |connection|
-      metrics_conn = Azure::Armrest::Insights::MetricsService.new(connection)
+      # TODO: code smell - should be in a 'collector' class instantiated here / find a better solution?
+      @connection = connection
```
agrare: Hm making connection an instance var when it is in a block feels wrong, and it looks like it could be solved by just passing it in to the metrics_url/definition_url methods

app/models/manageiq/providers/azure/cloud_manager/metrics_capture.rb
```diff
@@ -77,76 +77,135 @@ def perf_collect_metrics(interval_name, start_time = nil, end_time = nil)
...
+      @start_time = start_time.iso8601
+      @end_time   = end_time.iso8601
```
agrare: Similar comment here, this already is an instance of the MetricsCapture class so idk why you'd need another collector class, but also having method args be instance vars to avoid passing arguments to another method feels wrong.

app/models/manageiq/providers/azure/cloud_manager/metrics_capture.rb
```diff
@@ -77,76 +77,135 @@ def perf_collect_metrics(interval_name, start_time = nil, end_time = nil)
...
+          ].join(' and ')
+
+          table_data, _timings = Benchmark.realtime_block(:capture_counters) do
+            storage_account.table_data(
+              table_info.table_name,
+              storage_account_key,
+              :filter => filter,
+              :select => 'TIMESTAMP,Average',
+              :all    => true
+            )
```
agrare: This is an interesting API, do they basically give you SQL level access to the metrics db over REST?

app/models/manageiq/providers/azure/cloud_manager/metrics_capture.rb
```diff
@@ -77,76 +77,135 @@ def perf_collect_metrics(interval_name, start_time = nil, end_time = nil)
...
+          t_start_time = parse_timestamp(table_info.start_time)
+          t_end_time   = parse_timestamp(table_info.end_time).end_of_day
+          next unless time_interval.overlaps?(t_start_time..t_end_time)
+
+          filter = [
+            "CounterName eq '#{metric_name}'",
+            "PartitionKey eq '#{metric_location.partition_key}'",
+            "Timestamp ge datetime'#{@start_time}'",
+            "Timestamp le datetime'#{@end_time}'",
+          ].join(' and ')
```
agrare: Much benefit of joining an array of essentially constant strings with some interpolation vs just one long string?  Did you break it up like this to avoid 125 char line length limit?

app/models/manageiq/providers/azure/cloud_manager/metrics_capture.rb
```diff
@@ -77,76 +77,135 @@ def perf_collect_metrics(interval_name, start_time = nil, end_time = nil)
...
 
-        next if timestamped_values.empty?
+      ## Organize data in the proper form
+
+      COUNTERS_INFO.values.flatten.each do |counter_info|
+        metric_values = metrics.values_at(*counter_info.metrics).compact
+        next if metric_values.empty?
+
+        # { timestamp => [value, ...], ... }
+        timestamped_values = metric_values.each_with_object({}) do |ts_values, memo|
```
agrare: Just a style thing but could this be simplified to something like

```
timestamped_values = Hash.new { |h, k| h[k] = [] }
metric_values.each { |ts, values| timestamped_values[ts].concat(values) }
```

agrare: If metric_values is already a hash with the timestamp as the key then what exactly is this whole thing transforming it from and to?

app/models/manageiq/providers/azure/cloud_manager/metrics_capture.rb
agrare: Yeah agreed seems like overkill

spec/models/manageiq/providers/azure/cloud_manager/metrics_capture_spec.rb
```diff
@@ -4,58 +4,71 @@
...
+
+  context "handles errors" do
     let(:metric) { described_class.new(self) }
-    let(:armrest_service) { double('::Azure::Armrest::ArmrestService') }
 
     before do
       allow(self).to receive(:ext_management_system).and_return(ems)
       allow(self).to receive(:name).and_return('target1')
       allow(self).to receive(:resource_group).and_return(group)
+      allow(self).to receive(:id).and_return('1')
```
agrare: Could most of these stubs be replaced by a just a let and FactoryBot.create ?  Stubbing `:id` seems strange

spec/models/manageiq/providers/azure/cloud_manager/metrics_capture_spec.rb
AlexanderZagaynov: Dan, I don't think we should test `azure-armrest` gem's internals here. This line can be replaced with `OpenStruct.new(resource_url: 'something')` here with exactly the same result. It's just a stub.

app/models/manageiq/providers/azure/cloud_manager/metrics_capture.rb
AlexanderZagaynov: Adam, if you remember recent PR to the API, where I fixed similar timestamp format issue - it was caused by a lack of a common unified approach. That's why I made this method here. It is used in all the places where it is needed, and makes sure we'll get the same timestamp format everywhere.

app/models/manageiq/providers/azure/cloud_manager/metrics_capture.rb
AlexanderZagaynov: Honestly, I'm surprised by this question. Why do we hardcode gem versions everywhere? It is the same - every api version has it's own differences, and this code relies on the exact specific version.

app/models/manageiq/providers/azure/cloud_manager/metrics_capture.rb
AlexanderZagaynov: yes, passing shared variable in all the methods it is needed - similarly smelly approach. Other approach would be to create a class, where this variable will be passed on initialization, and it will be shared within this class as attribute. I didn't want to start such additional class making, because I see it as a part of the biggest refactoring I mentioned.

app/models/manageiq/providers/azure/cloud_manager/metrics_capture.rb
AlexanderZagaynov: because current class doesn't has this variables on initialization, they come here through `perf_collect_metrics` method, not via `initialize`.

app/models/manageiq/providers/azure/cloud_manager/metrics_capture.rb
AlexanderZagaynov: yes. problem is that the table names are unpredictable, and I found only one way to get them so far (same as it was in the previous version of this code file).

app/models/manageiq/providers/azure/cloud_manager/metrics_capture.rb
AlexanderZagaynov: I did that for readability reasons. It is much easier to say we have 4 conditions here, that when reading a single line.

spec/models/manageiq/providers/azure/cloud_manager/metrics_capture_spec.rb
AlexanderZagaynov: as I wrote, I do not see this code as something perfect, it is just a stub to make old specs work.

app/models/manageiq/providers/azure/cloud_manager/metrics_capture.rb
agrare: You could make it a multiline string with a heredoc and not incur the runtime overhead of an array join (minimal sure, but unnecessary)

AlexanderZagaynov: are we getting into the tastes discussion again? you like your approach, I like mine, and this question price is such a nothing. how big should be performance win you proposing?

agrare: From the POV of a reviewer (all opinions my own) there are three types of taste discussion:

1. Personal Preference for which there is no significant upside or downside

An example of this might be if you really like using `class << self` vs `def self.` or you like `%i[a b]` instead of `[:a, :b]`

In this case I couldn't care less which you choose and really try to keep my personal preferences out of the review unless as a project we've standardized on one (e.g. %i[] instead of %i() was a deliberate change to the rubocop rules at the project level).

2. Keeping consistent with the current code vs refactoring

In this case while a different design might be objectively better, usually it is preferable to stay consistent with the rest of the method/file/codebase.

I believe this is the issue with making `@connection`, `@start_time`, and `@end_time` instance vars when they are currently intentionally method parameters.

Would it be better if the `MetricsCapture` class took `@start_time` and `@end_time` as instance params, probably.  And if we want to do that we should come up with the design in core and apply it to the providers and properly refactor it.

3. Cases where there don't appear to be any upsides with some downsides

That is what I believe this case is.  There is literally no readability improvement of this over
```
<<~QUERY
  CounterName eq '#{metric_name}' AND \
  PartitionKey eq '#{metric_location.partition_key}' AND \
  Timestamp ge datetime'#{@start_time}' AND \
  Timestamp le datetime'#{@end_time}'
QUERY
```

and what you have here might be your taste but is objectively worse (even if only slightly) with marginal to no benefits.

app/models/manageiq/providers/azure/cloud_manager/metrics_capture.rb
djberg96: I don't really expect people to tinker with the api-version strings in the settings, it's just a matter of having it in one place instead of (potentially) many places throughout the code for convenience. Also, MS will completely drop support for api versions after a while, but I guess we could just say this particular release is no longer supported then.

app/models/manageiq/providers/azure/cloud_manager/metrics_capture.rb
djberg96: So, it's more like an interface method then, ok. Maybe we could move this into the core class, but that's a separate discussion.

app/models/manageiq/providers/azure/cloud_manager/metrics_capture.rb
Ladas: @djberg96 so I am playing with this in Topo and I've decided to just use version `Latest`, since this is still tied to the gem version, correct? 

So the Latest will keep pointing to the same version, until we update the gem. And i'd expect they don't add new versions with patch release?

djberg96: @Ladas that is essentially the default behavior of the azure-armrest gem, but we lock in api-versions in ManageIQ because sometimes MS breaks things between versions.

Ladas: @djberg96 right, my point was: isn't it enough to specify gem version, which contains only specific API versions (as classes)? (we'd have to move to latest in the gem though)

Also I am using the Azure's official gem in topo, so not sure how arm-rest differs there.

djberg96: No, the azure-armrest gem itself is not tied to specific api-version strings (in most cases). It will generally try to use the latest, or latest preview:

https://github.com/ManageIQ/azure-armrest/blob/master/lib/azure/armrest/configuration.rb#L266-L286

djberg96: See also: https://github.com/ManageIQ/azure-armrest/wiki/Locking-Down-the-%22api-version%22-Strings

app/models/manageiq/providers/azure/cloud_manager/metrics_capture.rb
AlexanderZagaynov: maybe some stub into core, because each provider can have own timestamp format - which is something for subclassing. but yeah, it should be a part of a bigger picture

app/models/manageiq/providers/azure/cloud_manager/metrics_capture.rb
AlexanderZagaynov: Version, mentioned here - is the most recent, which supports table data info. Previously it was: https://github.com/ManageIQ/azure-armrest/blob/42d8728294643757d7f86f208a9a78100eae349f/lib/azure/armrest/insights/metrics_service.rb#L129
Anyway, I see all this code as an intermediate step on the road to Azure-own gemset instead of `azure-armrest`.

Ladas: Ah ok, the official gem has actual versioned classes in the gem.

app/models/manageiq/providers/azure/cloud_manager/metrics_capture.rb
AlexanderZagaynov: it transforms this:
```
{
  metric_name_1 => {
    timestamp1 => [value1, value2, ...],
    timestamp2 => ...
  },
  metric_name_2 => {
    timestamp1 => [value3, value4, ...],
    timestamp3 => ...
  },
  ...
}
```
into this:
```
{
  timestamp1 => [value1, value2, value3, value4],
  timestamp2 => ...
  ...
}
```

app/models/manageiq/providers/azure/cloud_manager/metrics_capture.rb
AlexanderZagaynov: done

agrare:
Thanks @AlexanderZagaynov !

simaishi:
Ivanchuk backport details:

```
$ git log -1
commit 8473d9e5d7db37de50d50071c612583bbdec5947
Date:   Mon Aug 12 11:59:14 2019 -0400

    Merge pull request #341 from AlexanderZagaynov/BZ-1671580_azure_metrics
    
    capture memory metric
    
    (cherry picked from commit 3a0964bc18bd8faa4ff438cc02ed948959d88cf2)
    
    Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1724415
```

PR #343, opened 2019-08-09
Set the vm_uid_ems in the event parser if it's a VM

djberg96:
At the moment the Azure provider event parser loses track of its original VM in the case of a delete event. This affects timeline information.

The core `EmsEvent` class relies on the `vm_uid_ems` value to be set in order to preserve the original VM information:

https://github.com/ManageIQ/manageiq/blob/master/app/models/ems_event.rb#L87

Since the `vm_ems_ref` and `vm_uid_ems` are identical in the Azure provider, the solution was to simply assign the same value. In addition to the specs, I tried creating and deleting a couple VM's in the Azure dev environment and, after a few minutes, I could see the delete event with the `vm_or_template_id` and `vm_name` fields both set as expected.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1733384

Thanks go to @agrare for providing the solution.

### DIFF
--- a/app/models/manageiq/providers/azure/cloud_manager/event_parser.rb
+++ b/app/models/manageiq/providers/azure/cloud_manager/event_parser.rb
@@ -18,7 +18,7 @@ module ManageIQ::Providers::Azure::CloudManager::EventParser
     }
 
     resource_type = event["resourceType"]["value"].to_s.downcase
-    event_hash[:vm_ems_ref] = parse_vm_ref(event) if resource_type == INSTANCE_TYPE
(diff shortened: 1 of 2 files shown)


djberg96:
@miq-bot add_label bug, hammer/yes, ivanchuk/yes

simaishi:
Ivanchuk backport details:

```
$ git log -1
commit 4bd7f16ca57774cde18a1315d820ef25f6d87797
Date:   Fri Aug 9 10:02:42 2019 -0400

    Set the vm_uid_ems in the event parser if it's a VM.
    
    Add check for vm_uid_ems in event parser specs.
    
    (cherry picked from commit fe5dc6364b7f98477d77cfc06b78fd0a3ed28ae1)
    
    Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1733384
```

PR #347, opened 2019-09-03
fix invalid and syntax

AlexanderZagaynov:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1747858

### DIFF
--- a/app/models/manageiq/providers/azure/cloud_manager/metrics_capture.rb
+++ b/app/models/manageiq/providers/azure/cloud_manager/metrics_capture.rb
@@ -209,8 +209,8 @@ class ManageIQ::Providers::Azure::CloudManager::MetricsCapture < ManageIQ::Provi
     counter_values_by_mor[ems_ref] = counter_values
 
     return counters_by_mor, counter_values_by_mor
-  rescue ::Azure::Armrest::BadRequestException # Probably means region is not supported
-    msg = "Problem collecting metrics for #{resource_description}. "\
+  rescue ::Azure::Armrest::BadRequestException => err # Probably means region is not supported
+    msg = "Problem collecting metrics for #{resource_description}: #{err}. "\
           "Region [#{provider_region}] may not be supported."
     _log.warn(msg)
     return counters_by_mor, counter_values_by_mor
--- a/app/models/manageiq/providers/azure/cloud_manager/metrics_capture.rb
+++ b/app/models/manageiq/providers/azure/cloud_manager/metrics_capture.rb
@@ -140,9 +140,9 @@ class ManageIQ::Providers::Azure::CloudManager::MetricsCapture < ManageIQ::Provi
         storage_account_key  = storage_account_keys.fetch('key1')
 
         filter = <<~FILTER
-          CounterName eq '#{metric_name}' AND \
-          PartitionKey eq '#{metric_location.partition_key}' AND \
-          Timestamp ge datetime'#{start_time.iso8601}' AND \
+          CounterName eq '#{metric_name}' and \
+          PartitionKey eq '#{metric_location.partition_key}' and \
+          Timestamp ge datetime'#{start_time.iso8601}' and \
           Timestamp le datetime'#{end_time.iso8601}'
         FILTER
 


AlexanderZagaynov:
@agrare please review

app/models/manageiq/providers/azure/cloud_manager/metrics_capture.rb
```diff
@@ -209,8 +209,8 @@ def perf_collect_metrics(interval_name, start_time = nil, end_time = nil)
     counter_values_by_mor[ems_ref] = counter_values
 
     return counters_by_mor, counter_values_by_mor
-  rescue ::Azure::Armrest::BadRequestException # Probably means region is not supported
-    msg = "Problem collecting metrics for #{resource_description}. "\
+  rescue ::Azure::Armrest::BadRequestException => err # Probably means region is not supported
+    msg = "Problem collecting metrics for #{resource_description}: #{err}. "\
           "Region [#{provider_region}] may not be supported."
```
agrare: When the region isn't supported for metrics is that obvious from the exception message?  Wondering if we can drop the extra "may not be supported"

djberg96:
Changing "and" to "AND" fixes this? Wow, that is terrible.

agrare:
> Changing "and" to "AND" fixes this? Wow, that is terrible.

Yeah agreed

@AlexanderZagaynov from the BZ it sounded like only some VMs were impacted by this, the reporter said that windows vms worked but linux vms did not.  Is that accurate?  If so even more confused by changing `AND`

app/models/manageiq/providers/azure/cloud_manager/metrics_capture.rb
AlexanderZagaynov: I can't check that, don't have credentials for such region (tried all Germany, according to https://github.com/ManageIQ/manageiq-providers-azure/pull/302). Dan, maybe you can find and show us, what the error do they return in such case? @djberg96

agrare: I see @djberg96 said that `unfortunately the error message is not friendly.` so I'm good with this

AlexanderZagaynov:
@agrare it returns
```
[400] {"odata.error":{"code":"InvalidInput","message":{"lang":"en-US","value":"Syntax error at position 54 in 'CounterName eq '/builtin/memory/percentusedmemory' AND PartitionKey eq ':002Fsubscriptions:002Fd8dab060:002D0d39:002D4c77:002D82d7:002Dafc3ff5c7b06:002FresourceGroups:002Faz:002Drg:002Fproviders:002FMicrosoft:002ECompute:002FvirtualMachines:002Faz:002Dmetrics:002D1' AND Timestamp ge datetime'2019-09-03T09:46:17Z' AND Timestamp le datetime'2019-09-03T13:46:17Z''.\nRequestId:d154da70-a002-004b-0c5e-62caad000000\nTime:2019-09-03T13:47:02.7507975Z"}}} (cause: 400 Bad Request)
```
for linux instance with the big `AND`.
I'm testing it currently on windows, it behaves weirdly, investigating.

agrare:
> for linux instance with the big AND.  I'm testing it currently on windows, it behaves weirdly, investigating.

:scream: okay thanks

AlexanderZagaynov:
@agrare same error with the windows:
```
[400] {"odata.error":{"code":"InvalidInput","message":{"lang":"en-US","value":"Syntax error at position 53 in 'CounterName eq '\\Memory\\% Committed Bytes In Use' AND <<cut>>
```

@nachandr doesn't see it because she don't have `\Memory\% Committed Bytes In Use` metric in the list of available.
Nandini, please make sure you have the **OS guest diagnostics** enabled:
Screenshot from 2019-09-03 16-33-12

simaishi:
Ivanchuk backport details:

```
$ git log -1
commit 52e6b04be4e2da5d93e42e3418f53ec1342e9aa0
Date:   Tue Sep 3 13:31:08 2019 -0400

    Merge pull request #347 from AlexanderZagaynov/BZ-1747858_invalid_and_syntax
    
    fix invalid and syntax
    
    (cherry picked from commit 3eca30ccf94e9ebfa8ea409b80e0629140159bfc)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1724415
    https://bugzilla.redhat.com/show_bug.cgi?id=1747858
```

PR #350, opened 2019-09-06
Fix event timestamps for Azure

djberg96:
There's currently an issue where some events can get dropped in our event collector for Azure.

THE LONG EXPLANATION

The problem is that Azure event timestamps are not always when the event occurred, but when the event was generated by the Azure service processing the request. If there are multiple events happening in quick succession, like capture + generalize, the "capture" event could happen first, and then the "generalize" event happens later BUT might actually have a `submissionTimestamp` that's *before* the capture event. However, the `since` variable (the time controlling our query to the Azure event service) uses the later `eventTimestamp`, thus effectively skipping over the "backdated" generalize event.

In theory we should be using the field called `submissionTimestamp` instead. The problem is that Azure won't let us use it as a filter. You can *only* use `eventTimestamp`.

https://docs.microsoft.com/en-us/azure/azure-monitor/platform/activity-log-schema
https://docs.microsoft.com/en-us/rest/api/monitor/activitylogs/list

So, we need to change the logic to make sure these types of events don't get missed. ~~For an initial pass I'm changing it to the time of the last worker's heartbeat. I've also changed the default heartbeat to once every minute instead of once per 15 seconds to give the query time to actually complete before the next heartbeat.~~

Update: We've had to abandoned some of our original "simple" ideas because, to put it shortly, they don't work and events could still get skipped. So, the new strategy is as follows:

* Store the `dataEventId` as the `ems_ref` for the event. This is a unique ID for the event.
* Use the latest timestamp minus 2 minutes as our starting point instead of right at the most recent timestamp.
* Any overlap/duplicates will be automatically filtered out by the `EmsEvent` model. This required truncating the Azure `eventTimestamp` result to 6 digits to align with the underlying PostgreSQL table precision so that comparisons would work properly.
* Update specs, as well as the json files used by the specs.

While I was in here I made some other refactorings:

* Use the `operationName/localizedValue` as the default message string if the `description` field is blank (which it usually is).
* Use the `to_hash` method instead of `JSON.parse` because the underlying azure-armrest gem already stores a hash representation of the results internally.
* Use accessors instead of directly using instance variables.
* Add the `correlationId` to the collected data. This is essentially a group ID that may be useful.
* Added lots of comments.

Some answers to questions you may have:

Q: Why 2 minutes?
A: Because near as we can tell the `submissionTimestamp` and `eventTimestamp` have always been within 2 minutes of each other.

Q: Why not just use `submissionTimestamp`?
A: As I mentioned above, the Azure REST API will not let you filter on it.

~~Q: Isn't this extra query inefficient?
A: It does suck that we have to make an extra query to avoid duplicates, but keep in mind it's only going to look for those event id's that it found since the last heartbeat. Also keep in mind that both the `ems_ref` and `timestamp` fields in the `event_streams` table are indexed.~~

THE SHORT VERSION

Some Azure events are getting skipped because of the way Azure timestamps the events.

https://bugzilla.redhat.com/show_bug.cgi?id=1724312


app/models/manageiq/providers/azure/cloud_manager/event_catcher/stream.rb
```diff
@@ -29,13 +30,14 @@ def each_batch
...
-    fields = 'authorization,description,eventName,eventTimestamp,resourceGroupName,resourceProviderName,resourceId,resourceType'
+    fields = 'authorization,correlationId,description,eventId,eventName,eventTimestamp,resourceGroupName,resourceProviderName,resourceId,resourceType,submissionTimestamp'
     events = connection.list(:filter => filter, :select => fields, :all => true).sort_by(&:event_timestamp)
 
     # HACK: the Azure Insights API does not support the 'gt' (greater than relational operator)
     # therefore we have to poll from 1 millisecond past the timestamp of the last event to avoid
     # gathering the same event more than once.
-    @since = one_ms_from_last_timestamp(events) unless events.empty?
+    #@since = one_ms_from_last_timestamp(events) unless events.empty?
+    @since = @worker.last_heartbeat unless events.empty?
```
agrare: I could be wrong (cc @jrafanie) but I think the previous worker record will be destroyed by the time the next worker is started.  Testing this theory out locally.

djberg96: Ah, crud.

agrare: Yeah so I tried delivering a SIGKILL to my MiqEventHandler:

```
>> MiqEventHandler.first.id
=> 17

$ ps aux | grep MiqEventHandler | grep -v grep | awk '{print $2}' | xargs kill -KILL

... wait for MiqServer to restart it

>> MiqEventHandler.first.id
=> 18
```

agrare: @djberg96 I saw your comment from https://gitter.im/ManageIQ/manageiq?at=5d72a5c82f93ab5a15190ec3 and assumed by "last time it ran" as when the previous worker was running when the next worker starts up, is that what you want?

djberg96: @agrare That sounds right. How do I do that?

agrare: Okay, so that's an issue because of [:point_up:](https://github.com/ManageIQ/manageiq-providers-azure/pull/350/files#r321889178)

I don't know of a good way to tell when the previous worker instance was running.

Is that the issue here though?  The worker restarted and the first event that was delivered had a timestamp newer than the next event.

djberg96: @agrare on second thought, maybe that doesn't matter after all, as long as we're getting the latest time that the runner executed. We're tinkering with it now...

agrare: Why not make `@since` the last time when events were requested then?  I don't think you want last_heartbeat because that is relatively infrequent compared with the event catcher

djberg96: @agrare Yeah, we were thinking of just going with `Time.now - polling_interval`.

djberg96:
After lots of experimentation with John Dupuy's help, I think the unfortunate reality is that we're going to have to do some sort of post-processing that will require another request and/or database query.

Still tinkering...

app/models/manageiq/providers/azure/cloud_manager/event_catcher/stream.rb
```diff
@@ -26,40 +31,72 @@ def each_batch
...
+  # Go back a maximum of 1 hour if the newest record is older than that to avoid
+  # hitting the account's request limits. This should only happen in practice if
+  # the appliance has been down for a while and was restarted.
+  #
+  def most_recent_time
+    result = EventStream.select(:timestamp).where(:source => 'AZURE').order('timestamp desc').limit(1).first
 
-  def one_ms_from_last_timestamp_as_f(events)
-    Time.zone.parse(events.last.event_timestamp).to_f + 0.001
+    if result
```
AlexanderZagaynov:
```ruby
timestamp ? format_timestamp([1.hour.ago, timestamp - 2.minutes].max) : startup_interval
```

app/models/manageiq/providers/azure/cloud_manager/event_parser.rb
```diff
@@ -11,10 +11,11 @@ def self.event_to_hash(event, ems_id)
     event_hash = {
       :source     => "AZURE",
       :timestamp  => event["eventTimestamp"],
-      :message    => event["description"].presence,
+      :message    => event["description"].presence || event["operationName"]["localizedValue"].presence,
```
AlexanderZagaynov: What do you think about `event.dig(*%w[operationName localizedValue]).presence`?
And you'll get the free insurance against potential `nil[]` call error: https://devdocs.io/ruby~2.4/hash#method-i-dig

app/models/manageiq/providers/azure/cloud_manager/event_parser.rb
```diff
@@ -11,10 +11,11 @@ def self.event_to_hash(event, ems_id)
...
       :ems_id     => ems_id,
       :event_type => event_type,
-      :full_data  => event
+      :full_data  => event,
+      :ems_ref    => event["eventDataId"]
```
AlexanderZagaynov: It might be the matter of taste, but I'm always trying to have a comma on the such last line (ruby allows that), because it will be looking nicer on the diff next time somebody else will add some new line. You can see as example here line above marked as changed because of that comma.

app/models/manageiq/providers/azure/cloud_manager/event_catcher/stream.rb
```diff
@@ -1,4 +1,18 @@
...
+    description
+    eventDataId
+    eventName
+    eventTimestamp
+    operationName
+    resourceGroupName
+    resourceProviderName
+    resourceId
+    resourceType
+  ].join(',')
```
AlexanderZagaynov: I would add `.freeze` here, as it's a constant.

app/models/manageiq/providers/azure/cloud_manager/event_catcher/stream.rb
```diff
@@ -25,7 +39,7 @@ def stop
   #
   def each_batch
     while @collecting_events
-      yield get_events.collect { |e| JSON.parse(e) }
+      yield get_events.collect { |e| e.to_hash }
```
AlexanderZagaynov: Why not go further? `get_events.collect(&:to_hash)`

djberg96: Done!

app/models/manageiq/providers/azure/cloud_manager/event_catcher/stream.rb
djberg96: Done!

app/models/manageiq/providers/azure/cloud_manager/event_parser.rb
djberg96: I think it's a habit from the past. If I remember correctly, older versions of Ruby would choke on trailing commas in lists.

app/models/manageiq/providers/azure/cloud_manager/event_parser.rb
djberg96: I don't think it's possible to get a nil call error if we explicitly select `operationName`, since it's part of the output schema, though I guess it couldn't hurt.

app/models/manageiq/providers/azure/cloud_manager/event_catcher/stream.rb
```diff
@@ -26,40 +31,72 @@ def each_batch
...
-  end
+  # Retrieve the most recent Azure event minus 2 minutes, or the startup interval
+  # if no records are found.
+  #
+  # Go back a maximum of 1 hour if the newest record is older than that to avoid
+  # hitting the account's request limits. This should only happen in practice if
+  # the appliance has been down for a while and was restarted.
+  #
+  def most_recent_time
+    result = EventStream.select(:timestamp).where(:source => 'AZURE').order('timestamp desc').limit(1).first
```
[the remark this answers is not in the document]
djberg96: Ok, updated.

Not sure what you mean by filtering by ems id. We want the most recent time of the latest Azure event.

app/models/manageiq/providers/azure/cloud_manager/event_catcher/stream.rb
```diff
@@ -26,40 +31,72 @@ def each_batch
...
-    filter = @since ? "eventTimestamp ge #{@since}" : "eventTimestamp ge #{startup_interval}"
-    fields = 'authorization,description,eventName,eventTimestamp,resourceGroupName,resourceProviderName,resourceId,resourceType'
-    events = connection.list(:filter => filter, :select => fields, :all => true).sort_by(&:event_timestamp)
+    filter = "eventTimestamp ge #{most_recent_time}"
+    fields = 'authorization,correlationId,description,eventDataId,eventName,eventTimestamp,operationName,resourceGroupName,resourceProviderName,resourceId,resourceType'
+
+    events = connection.list(:filter => filter, :select => fields, :all => true)
+
+    if events.present?
+      existing_records = EventStream.select(:ems_ref).where(:source => 'AZURE', :ems_ref => events.map(&:event_data_id)).map(&:ems_ref)
```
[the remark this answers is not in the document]
djberg96: Eh, seems like a trivial access, but I'll update it if necessary.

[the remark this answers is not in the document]
agrare: Hey @djberg96, @AlexanderZagaynov is correct we are trying to keep event catchers (as many provider workers as possible really) from hitting the database directly.

Every one of these that we add will make it harder for us to move these out of rails/active_record which we need to do to get our memory usage down.

app/models/manageiq/providers/azure/cloud_manager/event_catcher/stream.rb
```diff
@@ -1,65 +1,115 @@
...
-    filter = @since ? "eventTimestamp ge #{@since}" : "eventTimestamp ge #{startup_interval}"
-    fields = 'authorization,description,eventName,eventTimestamp,resourceGroupName,resourceProviderName,resourceId,resourceType'
-    events = connection.list(:filter => filter, :select => fields, :all => true).sort_by(&:event_timestamp)
+    filter = "eventTimestamp ge #{most_recent_time}"
+
+    events = connection.list(:filter => filter, :select => SELECT_FIELDS, :all => true)
+
+    if events.present?
+      existing_records = EventStream.select(:ems_ref).where(:source => 'AZURE', :ems_ref => events.map(&:event_data_id)).map(&:ems_ref)
+      events = events.reject{ |e| existing_records.include?(e.event_data_id) } if existing_records.present?
```
agrare: If you're just trying to prevent saving duplicate events we should be able to do that in the event handler.

There already is logic there to prevent duplicate events being created, https://github.com/ManageIQ/manageiq/blob/master/app/models/ems_event.rb#L188-L194

app/models/manageiq/providers/azure/cloud_manager/event_parser.rb
agrare: > because it will be looking nicer on the diff next time somebody else will add some new line

^ I do the same thing, rubocop started complaining about it recently but I hate seeing the previous line in a diff when you just add something after it.

Not sure if this is one of those new rubocop rules that just got added automatically or if it was a conscious decision

app/models/manageiq/providers/azure/cloud_manager/event_catcher/stream.rb
djberg96: @agrare Oh, how do we use that exactly? I see our Runner uses `EmsEvent.add_queue` at:

 https://github.com/ManageIQ/manageiq-providers-azure/blob/master/app/models/manageiq/providers/azure/cloud_manager/event_catcher/runner.rb#L29

Just change it to this?

`EmsEvent.add_queue('add', @cfg[:ems_id], EmsEvent.create(event_hash))`

agrare: It _should_ happen automatically, since `EmsEvent.add_queue` calls `EmsEvent.add` which in turn calls `.create_event`

djberg96: Oops, that's a private method, hmm...

agrare: @djberg96 were you seeing duplicate events get created or was this a "just in case" thing?

app/models/manageiq/providers/azure/cloud_manager/event_catcher/stream.rb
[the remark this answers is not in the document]
AlexanderZagaynov: Can you theoretically have two and more records for Azure provider? With different credentials of different departments of some organization, in the two different regions? If so, `event_streams` table will have records with different `ems_id`. You'll get the most recent timestamp from both of those emses, but it will only be valid for one of them.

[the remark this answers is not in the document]
AlexanderZagaynov: And we still have direct database access here...

app/models/manageiq/providers/azure/cloud_manager/event_catcher/stream.rb
```diff
@@ -1,65 +1,107 @@
...
+  # in quick succession. We must therefore begin our query a couple minutes
+  # back from our most recent timestamp, and filter out any duplicates.
+  #
+  # See https://bugzilla.redhat.com/show_bug.cgi?id=1724312 for details.
+  #
   def get_events
-    # Grab only events for the last minute if this is the first poll
-    filter = @since ? "eventTimestamp ge #{@since}" : "eventTimestamp ge #{startup_interval}"
-    fields = 'authorization,description,eventName,eventTimestamp,resourceGroupName,resourceProviderName,resourceId,resourceType'
-    events = connection.list(:filter => filter, :select => fields, :all => true).sort_by(&:event_timestamp)
```
AlexanderZagaynov: You lost `.sort_by(&:event_timestamp)`. Was it by intention?

djberg96: I guess I'll have to put it back. I originally didn't think it was necessary because I was going to use a sql query to get the latest timestamp anyway.

app/models/manageiq/providers/azure/cloud_manager/event_catcher/stream.rb
```diff
@@ -1,65 +1,107 @@
...
-  end
+  # Retrieve the most recent Azure event minus 2 minutes, or the startup interval
+  # if no records are found.
+  #
+  # Go back a maximum of 1 hour if the newest record is older than that to avoid
+  # hitting the account's request limits. This should only happen in practice if
+  # the appliance has been down for a while and was restarted.
+  #
+  def most_recent_time
+    result = EventStream.where(source: 'AZURE').maximum(:timestamp)
```
agrare: Aside from the problem that this hits the database

1. This doesn't scope by ems_id so it'll get _all_ azure events not just for this EMS
2. This is called by get_events _every time_ so it is doing db queries even though after the first call the event catcher has the most up to date info on the events
3. I don't think we need to try to get all events while the event_catcher was down.  If MiqServer shuts down the event_catcher we specifically spin up another one immediately without waiting for the old one to stop specifically for this reason.  And if the appliance was down then the appliance was down its okay to miss things.

What about something like this?
```
EVENT_TIMESTAMP_BUFFER = 2.minutes
attr_accessor :most_recent_time

def initialize(ems)
  ...
  self.most_recent_time = Time.now - EVENT_TIMESTAMP_BUFFER
end

def get_events
  ...
  self.most_recent_time = events.map(&:event_timestamp).sort.last - EVENT_TIMESTAMP_BUFFER
end
```

app/models/manageiq/providers/azure/cloud_manager/event_catcher/stream.rb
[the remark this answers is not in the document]
djberg96: As per a gitter discussion, it turns out we didn't need this at all because `EmsEvent.create_event` is supposed to be filtering out duplicates already. The problem was the timestamp precision from Azure is 7 digits, while Postgres uses 6 digits. The comparison would thus fail, and a duplicate would get added. So, the solution was to truncate the date to 6 digits.

app/models/manageiq/providers/azure/cloud_manager/event_parser.rb
djberg96: Ok, done.

app/models/manageiq/providers/azure/cloud_manager/event_parser.rb
djberg96: If the rule changes, or we want to override it locally, we can update it. But for now I'm just gonna go with Rubocop's recommendations.

CHANGELOG.md
```diff
@@ -5,7 +5,7 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 
 
-## Hammer-10
+## Ivanchuk-1
```
agrare: Bad rebase?

app/models/manageiq/providers/azure/cloud_manager/event_catcher/stream.rb
djberg96: As per our gitter conversation, this ultimately turned out to be an issue with timestamp precision and comparison.

app/models/manageiq/providers/azure/cloud_manager/event_catcher/stream.rb
```diff
@@ -1,67 +1,112 @@
...
-  end
+  # Retrieve the most recent Azure event minus 2 minutes, or the startup interval
+  # if no records are found.
+  #
+  # Go back a maximum of 1 hour if the newest record is older than that to avoid
+  # hitting the account's request limits. This should only happen in practice if
+  # the appliance has been down for a while and was restarted.
+  #
+  def most_recent_time
+    result = EventStream.where(:source => 'AZURE', :ems_id => ems.id).maximum(:timestamp)
```
agrare: I think you missed the point of https://github.com/ManageIQ/manageiq-providers-azure/pull/350#discussion_r326146682

app/models/manageiq/providers/azure/cloud_manager/event_catcher/stream.rb
djberg96: 1. Oops, fixed.
2. I'm not sure I follow. Won't a `sort_by` generate another Ruby array internally, every time? The database query, on the other hand, will have its explain plan cached after the first call for 1 or 2 minutes won't it? The explain plan looks pretty zippy to me.
3. Eh, it only goes back an hour at most the way I wrote it there, but if you say we don't need to worry about it then I guess I won't.

app/models/manageiq/providers/azure/cloud_manager/event_catcher/stream.rb
```diff
@@ -1,23 +1,28 @@
...
+  # Creates an event monitor. Used internally by the Runner.
   #
   def initialize(ems)
     @ems = ems
     @collecting_events = false
     @since = nil
   end
 
-  # Start capturing events
+  # Sets a boolean used by the +each_batch+ method that indicates
```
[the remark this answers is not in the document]
djberg96: Thanks!

CHANGELOG.md
djberg96: I dunno what the heck I did here.

djberg96: Alright, should be fixed.

app/models/manageiq/providers/azure/cloud_manager/event_catcher/stream.rb
agrare: > Won't a sort_by generate another Ruby array internally, every time?

`events.max_by(&:event_timestamp)` would prevent the new ruby array

> The database query, on the other hand, will have its explain plan cached after the first call for 1 or 2 minutes won't it? The explain plan looks pretty zippy to me.

Anything done in memory is going to be much faster than a database query, but performance isn't even my main concern here.  We're trying to avoid unnecessary database access by provider workers because we want to run them without db access at all in the (hopefully near) future and stuff like this will make that harder.

djberg96: Ok, will update.

app/models/manageiq/providers/azure/cloud_manager/event_catcher/stream.rb
djberg96: Well, I also wanted to test it quick. :)

app/models/manageiq/providers/azure/cloud_manager/event_catcher/stream.rb
AlexanderZagaynov: Honestly, I have no idea why it needed. Theoretically, it affects on the order of the new queue tasks.

app/models/manageiq/providers/azure/cloud_manager/event_catcher/stream.rb
AlexanderZagaynov: As an option, why not to store such intermediate values in the cache? We have memcache in use, which is shared same as the database, but should be faster.
```ruby
def last_timestamp_key
  @last_timestamp_key ||= "AzureEventsLastTimestamp/#{ems.id}"
end
...
[Rails.cache.fetch(last_timestamp_key) { 2.minutes.ago }, 1.hour.ago].max
...
Rails.cache.write(last_timestamp_key, last_timestamp)
```

app/models/manageiq/providers/azure/cloud_manager/event_catcher/stream.rb
djberg96: It gives us the jumping point for the next rest api query.

app/models/manageiq/providers/azure/cloud_manager/event_catcher/stream.rb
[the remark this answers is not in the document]
AlexanderZagaynov: > Not sure what you mean by filtering by ems id

ok, Adam explained better [here](https://github.com/ManageIQ/manageiq-providers-azure/pull/350#discussion_r326146682).

app/models/manageiq/providers/azure/cloud_manager/event_catcher/stream.rb
agrare: I think just an instance variable is sufficient

app/models/manageiq/providers/azure/cloud_manager/event_catcher/stream.rb
AlexanderZagaynov: btw, you can simply `ems: ems` instead of `ems_id: ems.id`, and you don't need a `source` filter with `ems`.
or even
`ems.ems_events.maximum(:timestamp)`

app/models/manageiq/providers/azure/cloud_manager/event_catcher/stream.rb
AlexanderZagaynov: Problem happens when we have worker restarted. That's why instance variable is not sufficient, and we have this database call. However, what I see, is that we have three possible values of timestamp query:
- `1.hour.ago` (max for the previous events timestamp from the database)
- `1.hour.ago...2.minutes.ago` (somewhere in the range, from the database)
- `2.minutes ago` (when worker starts and there is no events in the database)
What difference is here and what is the logic @djberg96 ?

agrare: > Problem happens when we have worker restarted.

If the worker is restarted while the appliance is up, we spin up a new event_catcher immediately so going back a few minutes will cover the overlap.  We might send some duplicate events but that's why the timestamp fix is so critical because we already handle this case in core.

djberg96: Ok, updated.

app/models/manageiq/providers/azure/cloud_manager/event_catcher/stream.rb
djberg96: Updated.

app/models/manageiq/providers/azure/cloud_manager/event_catcher/stream.rb
djberg96: As per gitter, we were getting duplicates because of timestamp comparison failures. Fixed now.

app/models/manageiq/providers/azure/cloud_manager/event_catcher/stream.rb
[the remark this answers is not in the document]
djberg96: Updated.

app/models/manageiq/providers/azure/cloud_manager/event_catcher/stream.rb
```diff
@@ -1,67 +1,107 @@
...
-    fields = 'authorization,description,eventName,eventTimestamp,resourceGroupName,resourceProviderName,resourceId,resourceType'
-    events = connection.list(:filter => filter, :select => fields, :all => true).sort_by(&:event_timestamp)
-
-    # HACK: the Azure Insights API does not support the 'gt' (greater than relational operator)
-    # therefore we have to poll from 1 millisecond past the timestamp of the last event to avoid
-    # gathering the same event more than once.
-    @since = one_ms_from_last_timestamp(events) unless events.empty?
+    filter = "eventTimestamp ge #{most_recent_time}"
+    events = connection.list(:filter => filter, :select => SELECT_FIELDS, :all => true)
+    since  = events.max_by(&:event_timestamp)
```
AlexanderZagaynov: `self.since =`, otherwise you'll set just a local variable within this method scope

AlexanderZagaynov: and `events.max_by(&:event_timestamp)&.event_timestamp` to have timestamp or `nil` as the result

app/models/manageiq/providers/azure/cloud_manager/event_catcher/stream.rb
djberg96: @AlexanderZagaynov It would actually have to be `:ext_management_system => ems`. :)

But, anyway, I ditched it.

app/models/manageiq/providers/azure/cloud_manager/event_catcher/stream.rb
agrare: :+1:

app/models/manageiq/providers/azure/cloud_manager/event_catcher/stream.rb
```diff
@@ -1,67 +1,107 @@
...
-    fields = 'authorization,description,eventName,eventTimestamp,resourceGroupName,resourceProviderName,resourceId,resourceType'
-    events = connection.list(:filter => filter, :select => fields, :all => true).sort_by(&:event_timestamp)
-
-    # HACK: the Azure Insights API does not support the 'gt' (greater than relational operator)
-    # therefore we have to poll from 1 millisecond past the timestamp of the last event to avoid
-    # gathering the same event more than once.
-    @since = one_ms_from_last_timestamp(events) unless events.empty?
+    filter = "eventTimestamp ge #{most_recent_time}"
+    events = connection.list(:filter => filter, :select => SELECT_FIELDS, :all => true)
+    self.since = events.max_by(&:event_timestamp)
```
AlexanderZagaynov: you still getting the object here, not timestamp
https://devdocs.io/ruby~2.4/enumerable#method-i-max_by

app/models/manageiq/providers/azure/cloud_manager/event_catcher/stream.rb
djberg96: Fixed the accessor, thanks.

app/models/manageiq/providers/azure/cloud_manager/event_catcher/stream.rb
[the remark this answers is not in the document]
djberg96: Updated.

app/models/manageiq/providers/azure/cloud_manager/event_catcher/stream.rb
djberg96: I guess we're ditching the 1.hour.ago bit...

app/models/manageiq/providers/azure/cloud_manager/event_catcher/stream.rb
```diff
@@ -1,67 +1,107 @@
...
-  def one_ms_from_last_timestamp_as_f(events)
-    Time.zone.parse(events.last.event_timestamp).to_f + 0.001
+  # Retrieve the most recent Azure event minus the timestamp buffer, or the
+  # startup interval if no records are found.
+  #
+  def most_recent_time
+    if since
+      format_timestamp(since - EVENT_TIMESTAMP_BUFFER)
+    else
+      startup_interval
```
AlexanderZagaynov: maybe just put this method contents here, to reduce amount of one-use methods?

app/models/manageiq/providers/azure/cloud_manager/event_catcher/stream.rb
djberg96: Updated.

app/models/manageiq/providers/azure/cloud_manager/event_catcher/stream.rb
djberg96: Updated, thanks.

app/models/manageiq/providers/azure/cloud_manager/event_catcher/stream.rb
djberg96: Was getting around to it. :)

Done.

app/models/manageiq/providers/azure/cloud_manager/event_catcher/stream.rb
```diff
@@ -1,67 +1,100 @@
...
-    format_timestamp(time)
-  end
-
-  def one_ms_from_last_timestamp_as_f(events)
-    Time.zone.parse(events.last.event_timestamp).to_f + 0.001
+  # Retrieve the most recent Azure event minus the timestamp buffer, or the
+  # startup interval (2 minutes) if no records are found.
+  #
+  def most_recent_time
+    if since
```
AlexanderZagaynov: maybe one-liner?
```ruby
format_timestamp(since ? since - EVENT_TIMESTAMP_BUFFER : 2.minutes.ago)
```

djberg96:
@miq-bot remove_label wip

djberg96:
@miq-bot add_label bug, hammer/yes, ivanchuk/yes

djberg96:
@miq-bot add_label changelog/yes

simaishi:
Hammer backport details:

```
$ git log -1
commit efc7b57d34b34cc417fdf5782475b0348ded4afd
Date:   Thu Sep 19 15:42:38 2019 -0400

    Merge pull request #350 from djberg96/event_timestamps
    
    Fix event timestamps for Azure
    
    (cherry picked from commit a5a223857b4fcbd175b5d7bd51c2c3bfee998604)
    
    Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1733383
```

simaishi:
Ivanchuk backport details:

```
$ git log -1
commit 405b75262ed4cda1fed39a3be2e4c6b98d86e47b
Date:   Thu Sep 19 15:42:38 2019 -0400

    Merge pull request #350 from djberg96/event_timestamps
    
    Fix event timestamps for Azure
    
    (cherry picked from commit a5a223857b4fcbd175b5d7bd51c2c3bfee998604)
    
    Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1767784
```

PR #355, opened 2019-09-30
Parse the Azure Event Timestamp into a DateTime

agrare:
The Azure event_timestamp is returned as a string which we then attempt
to do arithmatic on to get a relative time for subsequent queries.

We have to parse the timestamp into a DateTime object first before we
can do `- 2.minutes`

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1756984


djberg96:
Dang, how was this working for us before? Hm...

Anyway, good catch. :)

agrare:
@djberg96 I'm going to write some specs for this so hold off on merging please :)

agrare:
We can add a lot more tests but this at least exercises this new code path

agrare:
@djberg96 can you take a look?

agrare:
> Dang, how was this working for us before? Hm...

I suspect the first get_events call worked, the second threw an exception and restarted the event catcher, which then did another first get_events call

PR #368, opened 2019-11-14
Rely on the default implementation of .ems_class

carbonin:
Requires https://github.com/ManageIQ/manageiq/pull/19516


carbonin:
These failures will be resolved once the PR to manageiq core is merged. See: https://github.com/ManageIQ/manageiq-cross_repo-tests/pull/12

PR #369, opened 2019-11-15
metrics_capture passes ems

kbrock:
This is the new api

rest of the build depends upon https://github.com/ManageIQ/manageiq/pull/19522


jrafanie:
@kbrock https://github.com/ManageIQ/manageiq/pull/19522 is merged, please restart this build

PR #373, opened 2019-12-06
Subclass VmScan job as ManageIQ::Providers::Azure::CloudManager::Scanning::Job

chessbyte:
State Machine diagram for this subclass of VmScan.

manageiq-providers-azure-cloud_manager-scanning-job

### DIFF
--- a/app/models/manageiq/providers/azure/cloud_manager/scanning/job.rb
+++ b/app/models/manageiq/providers/azure/cloud_manager/scanning/job.rb
@@ -1,0 +1,173 @@
+class ManageIQ::Providers::Azure::CloudManager::Scanning::Job < VmScan
+  # Make updates to default state machine to take into account snapshots
+  def load_transitions
+    super.tap do |transitions|
+      transitions.merge!(
+        :start_snapshot     => {'before_scan'               => 'snapshot_create'},
+        :snapshot_complete  => {'snapshot_create'           => 'scanning',
+                                'snapshot_delete'           => 'synchronizing'},
+        :snapshot_delete    => {'scanning'                  => 'snapshot_delete'},
+        :data               => {'snapshot_create'           => 'scanning',
+                                'scanning'                  => 'scanning',
+                                'snapshot_delete'           => 'snapshot_delete',
+                                'synchronizing'             => 'synchronizing',
+                                'finished'                  => 'finished'}
+      )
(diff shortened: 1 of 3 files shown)


app/models/manageiq/providers/azure/cloud_manager/scanning/job.rb
```diff
@@ -0,0 +1,173 @@
...
+      if vm.ext_management_system
+        _log.info("Deleting snapshot: reference: [#{mor}]")
+        begin
+          delete_snapshot(mor, vm)
+        rescue => err
+          _log.error(err.to_s)
+          return
+        rescue Timeout::Error
+          msg = "Request to delete snapshot timed out"
+          _log.error(msg)
```
djberg96: Do we want to return here on a timeout error? Otherwise the rest of the method will fire, and I worry it will inadvertently look like a success.

app/models/manageiq/providers/azure/cloud_manager/scanning/job.rb
```diff
@@ -0,0 +1,173 @@
...
+            when :smartProxy, :skipped then "Request to log snapshot user event with EMS timed out."
+            else "Request to create snapshot timed out"
+            end
+      _log.error(msg)
+      signal(:abort, msg, "error")
+    end
+  end
+
+  def call_snapshot_delete
+    _log.info("Enter")
```
djberg96: Do we want to include the method name in the log entry here?

chessbyte:
@roliveri @jerryk55 if you have the environment to test SmartState for Azure with this and the [base](https://github.com/ManageIQ/manageiq/pull/19607) PR, it would be much appreciated.

app/models/manageiq/providers/azure/cloud_manager/scanning/job.rb
chessbyte: @djberg96 while I agree, I would do that in a separate PR. I am just refactoring here - trying to change the least amount possible and ensure everything still works.

app/models/manageiq/providers/azure/cloud_manager/scanning/job.rb
chessbyte: @djberg96 I wonder how this worked prior to the refactoring.

PR #376, opened 2020-01-15
Switch metric for disk data

djberg96:
As per discussions with Nandini Chandra, I think we should switch the Azure metrics from "Per Disk Read/Write Bytes per sec" to just "Disk Read/Write Bytes". Partly this is because the current metrics doesn't really seem to collect much useful data, and partly because it's not even an option in the Azure portal's UI, and so isn't what a user would probably expect.

Addresses https://bugzilla.redhat.com/show_bug.cgi?id=1780706

### DIFF
--- a/app/models/manageiq/providers/azure/cloud_manager/metrics_capture.rb
+++ b/app/models/manageiq/providers/azure/cloud_manager/metrics_capture.rb
@@ -34,8 +34,10 @@ class ManageIQ::Providers::Azure::CloudManager::MetricsCapture < ManageIQ::Provi
       :counter_key => 'disk_usage_rate_average', # TODO: should be 'disk_usage_absolute_average',
       :source      => :api,
       :metrics     => [
-        'Disk Read Bytes',
-        'Disk Write Bytes',
+        'OS Disk Read Bytes/sec',
+        'OS Disk Write Bytes/sec',
+        'Data Disk Read Bytes/sec',
+        'Data Disk Write Bytes/sec',
       ].freeze,
       :calculation => ->(stats) { stats.sum / 1.kilobyte },
       :unit_key    => 'kilobytespersecond',
--- a/app/models/manageiq/providers/azure/cloud_manager/metrics_capture.rb
+++ b/app/models/manageiq/providers/azure/cloud_manager/metrics_capture.rb
@@ -34,8 +34,8 @@ class ManageIQ::Providers::Azure::CloudManager::MetricsCapture < ManageIQ::Provi
       :counter_key => 'disk_usage_rate_average', # TODO: should be 'disk_usage_absolute_average',
       :source      => :api,
       :metrics     => [
-        'Per Disk Read Bytes/sec',
-        'Per Disk Write Bytes/sec',
+        'Disk Read Bytes',
+        'Disk Write Bytes',
       ].freeze,
       :calculation => ->(stats) { stats.sum / 1.kilobyte },
       :unit_key    => 'kilobytespersecond',
--- a/spec/models/manageiq/providers/azure/cloud_manager/metrics_capture_spec.rb
+++ b/spec/models/manageiq/providers/azure/cloud_manager/metrics_capture_spec.rb
@@ -137,9 +137,8 @@ describe ManageIQ::Providers::Azure::CloudManager::MetricsCapture do
       expect(metrics_by_id_and_ts).to eq(expected_metrics)
     end
 
-    it "parses and aggregates read and write on disk" do
-      counters = stage_counter_data(['Per Disk Read Bytes/sec',
-                                     'Per Disk Write Bytes/sec'])
+    it "parses and aggregates read and write on disk", :disk do
+      counters = stage_counter_data(['Disk Read Bytes', 'Disk Write Bytes'])
 
       metric_data = [
         build_metric_data(982_252_000, "2016-07-23T07:20:00.5580968Z"),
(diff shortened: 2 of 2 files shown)


app/models/manageiq/providers/azure/cloud_manager/metrics_capture.rb
```diff
@@ -34,8 +34,8 @@ class ManageIQ::Providers::Azure::CloudManager::MetricsCapture < ManageIQ::Provi
       :counter_key => 'disk_usage_rate_average', # TODO: should be 'disk_usage_absolute_average',
       :source      => :api,
       :metrics     => [
-        'Per Disk Read Bytes/sec',
-        'Per Disk Write Bytes/sec',
+        'Disk Read Bytes',
+        'Disk Write Bytes',
```
agrare: @djberg96 since this is a rate (`disk_usage_rate_average`) I don't think that going to an absolute (Bytes) makes sense here.  How about `Disk Read Bytes/Sec` instead? (ref https://docs.microsoft.com/en-us/azure/azure-monitor/platform/metrics-supported#microsoftclassiccomputevirtualmachines)

djberg96: @agrare That's classic compute (i.e. the "old" azure). The current metrics list is:
```
- "Percentage CPU",
- "Network In",
- "Network Out",
- "Disk Read Bytes",
- "Disk Write Bytes",
- "Disk Read Operations/Sec",
- "Disk Write Operations/Sec",
- "CPU Credits Remaining",
- "CPU Credits Consumed",
- "Per Disk Read Bytes/sec",
- "Per Disk Write Bytes/sec",
- "Per Disk Read Operations/Sec",
- "Per Disk Write Operations/Sec",
- "Per Disk QD",
- "OS Per Disk Read Bytes/sec",
- "OS Per Disk Write Bytes/sec",
- "OS Per Disk Read Operations/Sec",
- "OS Per Disk Write Operations/Sec",
- "OS Per Disk QD",
- "Data Disk Read Bytes/sec",
- "Data Disk Write Bytes/sec",
-  "Data Disk Read Operations/Sec",
- "Data Disk Write Operations/Sec",
- "Data Disk Queue Depth",
- "OS Disk Read Bytes/sec",
- "OS Disk Write Bytes/sec",
- "OS Disk Read Operations/Sec",
- "OS Disk Write Operations/Sec",
- "OS Disk Queue Depth",
- "Inbound Flows",
- "Outbound Flows",
- "Inbound Flows Maximum Creation Rate",
- "Outbound Flows Maximum Creation Rate",
- "Premium Data Disk Cache Read Hit",
- "Premium Data Disk Cache Read Miss",
- "Premium OS Disk Cache Read Hit",
- "Premium OS Disk Cache Read Miss",
- "Network In Total",
- "Network Out Total"
```

djberg96: https://docs.microsoft.com/en-us/azure/azure-monitor/platform/metrics-supported#microsoftcomputevirtualmachines

agrare: Okay we still can't stick an absolute value in a rate column though, it looks like they split it up into "(OS Disk|Data Disk) Read Bytes/sec"?

djberg96: @agrare Ok, updated, though now I'm getting a spec failure that I'm having trouble nailing down. It seems to not use the staged data if there's more than 2 metrics.

agrare:
@djberg96 looks good, just want to confirm that these metrics are actually populated (unlike the previous metric)

djberg96:
@agrare I'm seeing a little bit, but not much.

agrare:
> I'm seeing a little bit, but not much.

What does that mean?  Like not all VMs have it, or it isn't as much IO/sec as you were expecting?

djberg96:
@agrare Just not as much as I was expecting, I guess I need a way to artificially generate some activity on the VM. The Azure portal shows the same general lack of activity as well, so it's not our code.

app/models/manageiq/providers/azure/cloud_manager/metrics_capture.rb
agrare: @djberg96 I think the values returned are just doubled whatever the spec is expecting since it is just summing all the stubbed counters

agrare:
I want to make sure that we don't have the same problem as:

> Partly this is because the current metrics doesn't really seem to collect much useful data

I assume that "doesn't seem to collect much useful data" means that `Per Disk Read Bytes/sec` wasn't being populated

djberg96:
Yeah, if this isn't collecting much useful data, what should we do instead (if anything)? Part of it is that I'm not totally certain what the UI is expecting, or if we can have multiple types of disk metrics.

agrare:
> Yeah, if this isn't collecting much useful data

I guess I'm confused then.  Is the problem that this isn't collecting much useful data, or that the VMs aren't doing much IO to the disks?  Aka is the metric we're collecting wrong?

> I'm not totally certain what the UI is expecting, or if we can have multiple types of disk metrics.

We only have one disk metric right now and it is the io rate aggregate for all disks.  We could add other metrics but it would require a schema change to add the columns to the metrics tables.

djberg96:
I don't think what we're collecting is necessarily wrong, I just think the VM's being checked aren't doing a heck of a lot. Which could also be the problem with the original metric, though there's still the issue of that metric not even being presented in the Azure Portal UI.

The plain "Disk Read/Write Bytes" does tend to show more activity, but as you say, it's not really appropriate.

agrare:
Okay, lets find out if we need this PR at all then.  If you're seeing larger numbers in `Disk Read/Write Bytes` than the `Per Sec` ones that's probably just because that is the total number of bytes since the VM was powered on if I had to guess

djberg96:
Going to close this for now.

PR #379, opened 2020-02-11
Add type to the list of validation dependencies

skateman:
This makes the form a little bit more semantically correct.

@miq-bot assign @djberg96


djberg96:
`This makes the form a little bit more semantically correct.` - How so? I guess I'm not following what it does in practice.

skateman:
@djberg96 well, the validation actually depends on the provider's type, even if it's obviously azure in the given context.

skateman:
@miq-bot assign @agrare

PR #380, opened 2020-02-25
Implement create/edit for DDF with adjusted schema and validation

skateman:
The DDF schema had to be adjusted to the DB schema field names, then the validation has been adjusted to pass again. On top of this I implemented the create and edit methods that can consume the data format produced by the forms.

Parent issue: https://github.com/ManageIQ/manageiq/issues/18818

@miq-bot add_reviewer @djberg96 
@miq-bot add_reviewer @agrare

### DIFF
--- a/app/models/manageiq/providers/azure/manager_mixin.rb
+++ b/app/models/manageiq/providers/azure/manager_mixin.rb
@@ -14,6 +14,20 @@ module ManageIQ::Providers::Azure::ManagerMixin
     connect(options)
   end
 
+  def edit_with_params(params)
+    default_endpoint = params.delete("endpoints").dig("default")
+    default_authentication = params.delete("authentications").dig("default")
+
+    tap do |ems|
+      ems.default_authentication.assign_attributes(default_authentication)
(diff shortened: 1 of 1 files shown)


app/models/manageiq/providers/azure/manager_mixin.rb
```diff
@@ -14,6 +14,20 @@ def verify_credentials(_auth_type = nil, options = {})
     connect(options)
   end
 
+  def edit_with_params(params)
+    default_endpoint = params.delete("endpoints").dig("default")
+    default_authentication = params.delete("authentications").dig("default")
+
+    tap do |ems|
+      ems.default_authentication.assign_attributes(default_authentication)
```
Fryguy: Out of curiosity, why `assign_attributes`?... elsewhere in code we use `attributes=`

skateman: Well, the argument is just a subset of attributes, not all of them. I'm not sure how `attributes=˙ handles the argument if it's just a subset :confused: 

Also I'm using it already in the amazon provider :trollface:

djberg96:
👍

PR #381, opened 2020-03-10
Fix shadowed exceptions in Scanning::Job

djberg96:
Codeclimate spotted these shadowed exceptions. The `TimeoutError` blocks will currently never be called since they're a subclass of `StandardError`. To see the current logic in action, run this:

```
require 'timeout'

begin
  raise Timeout::Error
rescue => err
  puts "REGULAR ERROR: #{err}"
rescue Timeout::Error => err
  puts "TIMEOUT ERROR: #{err}"
end

# RESULT => "REGULAR ERROR: Timeout::Error" # oops!
```


djberg96:
@miq-bot add_label bug

djberg96:
@jerryk55 or @roliveri is this your area of expertise? If so, please take a look.

PR #384, opened 2020-03-23
remove inventory object refresh

d-m-u:
we don't need any of this cause of the graph refresh yay


d-m-u:
@miq-bot add_label cleanup

agrare:
@d-m-u actually looks like this fails travis...but travis didn't run for this pr?

PR #391, opened 2020-05-08
Add a shell script for adding a user

djberg96:
I get enough requests for creating Azure account credentials that I think it's worth having this script around both for myself, and for potential future admins.

It's pretty straightforward script that creates a resource group with the same name as the userid, then creates a service principal attached to that resource group.


lib/tasks_private/add_user.sh
```diff
@@ -0,0 +1,46 @@
...
+#
+#   add_user.sh some_user [eastus 12345]
+#
+
+username=$1
+region="${2:-eastus}"
+subscription_id="${3:-$AZURE_SUBSCRIPTION_ID}"
+
+if [ -z "$username" ]; then
+  echo "You must specify a username argument. This should match the user's Redhat userid."
```
agrare, line 20: > This should match the user's Redhat userid.

Don't think we should have this ^ upstream

lib/tasks_private/add_user.sh
```diff
@@ -0,0 +1,46 @@
+#!/bin/bash
+
+# This script adds a user to the Azure dev environment using their Redhat
```
agrare, line 3: > using their Redhat userid

Same

agrare:
@djberg96 if this isn't a rake task should it be under `lib/tasks_private` ?  Seems like something which should be in `./tools/`

lib/tasks_private/add_user.sh
djberg96: Changed "Redhat" to "company".

lib/tasks_private/add_user.sh
```diff
@@ -0,0 +1,46 @@
+#!/bin/bash
+
+# This script adds a user to the Azure dev environment using their company
+# userid. By default, it will create a resource group in the east US region
```
agrare, line 4: I think company userid could just be replaced by userid, it could easily be a personal account

lib/tasks_private/add_user.sh
```diff
@@ -0,0 +1,46 @@
...
+#
+#   add_user.sh some_user [eastus 12345]
+#
+
+username=$1
+region="${2:-eastus}"
+subscription_id="${3:-$AZURE_SUBSCRIPTION_ID}"
+
+if [ -z "$username" ]; then
+  echo "You must specify a username argument. This should match the user's company userid."
```
agrare, line 20: Same, just `"user's ID"` would be fine

agrare:
@djberg96 can you move this to `tools/` ?

Fryguy:
@djberg96 Curious why this is in tasks_private if it's a shell script and not a rake task?  I would expect rake tasks to live there, and everything else in a top-level tools directory (or the bin directory, maybe, if we want it to be executable from an installed-gem point of view)

agrare:
:point_up: https://github.com/ManageIQ/manageiq-providers-azure/pull/391#issuecomment-625843959

lib/tasks_private/add_user.sh
djberg96: Updated.

lib/tasks_private/add_user.sh
djberg96: Updated.

djberg96:
@agrare, @Fryguy Ok, moved to `tools`.

PR #395, opened 2020-05-14
Loosen azure-armrest dependency to 0.9

djberg96:
Version 0.10.0 of the azure-armrest gem includes an upgrade to rest-client-2.1.0, as well as azure-signature-0.3.0. It also updates a metrics collection method that was woefully out of date.

Update: Due to issues with other libraries conflicting with rest-client versions, I've changed this to just "0.9". Since I released 0.9.16 which loosens the rest-client dependency, this should work with either rest-client version.

### DIFF
--- a/manageiq-providers-azure.gemspec
+++ b/manageiq-providers-azure.gemspec
@@ -13,7 +13,7 @@ Gem::Specification.new do |s|
 
   s.files = Dir["{app,config,lib}/**/*"]
 
-  s.add_dependency "azure-armrest", "~>0.10"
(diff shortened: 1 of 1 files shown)


agrare:
@djberg96 looks like rest-client is locked down to `~> 2.0.0` in core so we'd have to either allow that version in azure-armrest or bump core to `~> 2.1`

djberg96:
@agrare Yep, see https://github.com/ManageIQ/manageiq/pull/20173 :)

PR #399, opened 2020-06-11
Use vmId for uid_ems for instances

djberg96:
Currently within the Azure provider the ems_ref and uid_ems are the same. This PR updates the code so that the uid_ems is set to the `vmId` property instead which is a guid string.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1846159

### DIFF
--- a/app/models/manageiq/providers/azure/inventory/parser/cloud_manager.rb
+++ b/app/models/manageiq/providers/azure/inventory/parser/cloud_manager.rb
@@ -78,7 +78,7 @@ class ManageIQ::Providers::Azure::Inventory::Parser::CloudManager < ManageIQ::Pr
       next if (status = collector.power_status(instance)).blank?
 
       persister_instance = persister.vms.build(
-        :uid_ems             => uid,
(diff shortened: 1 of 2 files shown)


agrare:
> Addresses https://bugzilla.redhat.com/show_bug.cgi?id=1846159

I think you need "Fixes bz_url" for cfme-bot to pick up on it

djberg96:
@agrare Oops. And it seems editing it after the fact doesn't work, either. Oh well, I'll update the BZ manually.

agrare:
@djberg96 make sure the commit message has it also

djberg96:
@agrare Already did it manually. Do I still need to update the commit message?

simaishi:
@djberg96 @agrare this change is safe to go to ivanchuk?

agrare:
Ehh I'm hesitant to backport a change to the value of a column.

simaishi:
@dmetzger57 ^

@agrare thank you

dmetzger57:
Hold off on the backport for now, discussing with support.

PR #403, opened 2020-07-06
[WIP] Combine cloud and network managers

agrare:
A full refresh of a cloud manager always queues a refresh of the network manager, we can simplify this by refreshing both together

https://github.com/ManageIQ/manageiq/issues/20301


djberg96:
@agrare I'm just a wee bit late reviewing this. Was going to say it looks good to me in principle, it seems something changed in the meantime to prevent a merge.

agrare:
Yeah I need to get back to this, I was having issues getting the targeted refresh specs passing which is why I shelved it temporarily.

agrare:
Done in https://github.com/ManageIQ/manageiq-providers-azure/pull/454

PR #404, opened 2020-07-15
Filter out cloud networks with no cloud subnets during provisioning

djberg96:
In Azure you cannot provision a VM with a vnet that doesn't have at least one subnet. So, this PR adjusts the provisioning screen so that it will filter out any virtual networks that don't have at least one attached subnet. 

Part of a solution to https://github.com/ManageIQ/manageiq-providers-azure/issues/375 (the other part being a modification on the automate side at https://github.com/ManageIQ/manageiq-content/pull/662).

### DIFF
--- a/app/models/manageiq/providers/azure/cloud_manager/provision_workflow.rb
+++ b/app/models/manageiq/providers/azure/cloud_manager/provision_workflow.rb
@@ -15,6 +15,19 @@ class ManageIQ::Providers::Azure::CloudManager::ProvisionWorkflow < ManageIQ::Pr
     resource_groups.each_with_object({}) { |rg, hash| hash[rg.id] = rg.name }
   end
 
+  def allowed_cloud_networks(_options = {})
+    source = load_ar_obj(get_source_vm)
+    ems = source.try(:ext_management_system)
+
+    if ems.present?
+      ems.cloud_networks.select{ |cn| cn.cloud_subnets.size > 0 }.each_with_object({}) do |cn, hash|
+        hash[cn.id] = "#{cn.name} (#{cn.cidr})"
+      end
+    else
+      {}
+    end
+  end
+
   def allowed_cloud_subnets(_options = {})
     src = resources_for_ui
     if (cn = CloudNetwork.find_by(:id => src[:cloud_network_id]))
--- a/app/models/manageiq/providers/azure/cloud_manager/provision_workflow.rb
+++ b/app/models/manageiq/providers/azure/cloud_manager/provision_workflow.rb
@@ -15,12 +15,14 @@ class ManageIQ::Providers::Azure::CloudManager::ProvisionWorkflow < ManageIQ::Pr
     resource_groups.each_with_object({}) { |rg, hash| hash[rg.id] = rg.name }
   end
 
+  # In addition to restricting cloud networks by availability zone, we also
+  # restrict cloud networks to those that have at least one subnet.
+  #
   def allowed_cloud_networks(_options = {})
-    source = load_ar_obj(get_source_vm)
-    ems = source.try(:ext_management_system)
+    subset = CloudNetwork.where(:id => super.keys)
 
-    if ems.present?
-      ems.cloud_networks.distinct.joins(:cloud_subnets).where.not(:cloud_subnets => {:id => nil}).each_with_object({}) do |cn, hash|
+    if subset.present?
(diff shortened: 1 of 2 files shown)


djberg96:
You might be thinking, "Why not just create a default subnet if one isn't found?". First, I'm not sure if that's expected behavior by our users or for ManageIQ in general. Second, while it would work using the API, it would still not work on the UI side without additional work.

In other words, on the provisioning screen, if you were to select a vnet with no subnets, the UI would have to make a call to generate the subnet so that it could then present it too you as a selectable option. That, to me, seemed like a lot of work for what should be a pretty rare situation in general. Keep in mind that this issue was found by an engineer, not a user in the wild.

app/models/manageiq/providers/azure/cloud_manager/provision_workflow.rb
```diff
@@ -15,6 +15,19 @@ def allowed_resource_groups(_options = {})
     resource_groups.each_with_object({}) { |rg, hash| hash[rg.id] = rg.name }
   end
 
+  def allowed_cloud_networks(_options = {})
+    source = load_ar_obj(get_source_vm)
+    ems = source.try(:ext_management_system)
+
+    if ems.present?
+      ems.cloud_networks.select{ |cn| cn.cloud_subnets.size > 0 }.each_with_object({}) do |cn, hash|
```
Fryguy, line 23: This is an N+1...can you use one of the suggestions we gave in gitter?

Fryguy:
Also please fix the rubocops.

app/models/manageiq/providers/azure/cloud_manager/provision_workflow.rb
djberg96: Updated.

Fryguy:
@kbrock Please review.

djberg96:
Looks like I broke the specs a bit, working on it.

djberg96:
@kbrock fixed

Fryguy:
LGTM, but @agrare can you take a look

agrare:
@djberg96 so this looks like it no longer takes the selected availability zone into account, was that intentional?  I think we would end up with a different problem where a user picks an AZ but we try to put the instance on a cloud_network not in that AZ.

In the core [`allowed_cloud_networks`](https://github.com/ManageIQ/manageiq/blob/master/app/models/manageiq/providers/cloud_manager/provision_workflow.rb#L29-L33) method it is doing [`allowed_ci(:cloud_network, [:availability_zone], targets.map(&:id))`](https://github.com/ManageIQ/manageiq/blob/master/app/models/manageiq/providers/cloud_manager/provision_workflow.rb#L32) which ends up calling [`availability_zone_to_cloud_network`](https://github.com/ManageIQ/manageiq/blob/master/app/models/manageiq/providers/cloud_manager/provision_workflow.rb#L107-L118) which will filter cloud_networks if an availability zone is selected.

I think a better approach would be to add a `filter_cloud_networks` method to core which you override here to limit the targets, something like
```
def allowed_cloud_networks(_options = {})
  return {} unless (src = provider_or_tenant_object)
  targets = get_targets_for_source(src, :cloud_filter, CloudNetwork, 'all_cloud_networks')
  filter_cloud_networks!(targets)
  allowed_ci(:cloud_network, [:availability_zone], targets.map(&:id))
end

def filter_cloud_networks!(_targets)
end
```

then here you can do this filtering based on the targets passed in.

djberg96:
@agrare Can't I just call `super`, then filter from there? This wasn't meant to require a core change.

agrare:
Yeah sure assuming allowed_ci returns objects that are useful to you

agrare:
Looks like it returns an `{id => name}` hash so yeah you could use that easily

djberg96:
@agrare Updated.

agrare:
@djberg96 can this go back to jansa?  Looks like a good small fix to a legit bug.

djberg96:
@agrare Should be fine, though I don't have a BZ for it.

simaishi:
Jansa backport details:

```
$ git log -1
commit 7451d272f78a7705f904e7a7f2fa48223f246c70
Date:   Mon Jul 20 11:04:31 2020 -0400

    Merge pull request #404 from djberg96/restricted_vnets

    Filter out cloud networks with no cloud subnets during provisioning

    (cherry picked from commit dbbf287ad83997ad3ef260937c3604b27d8f0598)
```

PR #411, opened 2020-09-17
Use the tag_mapper from the core Persister

agrare:
Depends on:
- [x] https://github.com/ManageIQ/manageiq/pull/20561


agrare:
Travis is green after close/reopen https://travis-ci.com/github/ManageIQ/manageiq-providers-azure/pull_requests looks like the old run is stuck in the Checks list for some reason

agrare:
Force push appears to clear that up

PR #416, opened 2020-10-14
Case sensitive support for label mapping

lpichler:
related to https://github.com/ManageIQ/manageiq/pull/20289  or (https://github.com/ManageIQ/manageiq/pull/20289#discussion_r504746433)
Described in second point [here](https://github.com/ManageIQ/manageiq-ui-classic/pull/7273#issuecomment-698995573)

@miq-bot assign @agrare

### DIFF
--- a/app/models/manageiq/providers/azure/inventory/persister.rb
+++ b/app/models/manageiq/providers/azure/inventory/persister.rb
@@ -23,4 +23,10 @@ class ManageIQ::Providers::Azure::Inventory::Persister < ManageIQ::Providers::In
 
     initialize_inventory_collections
   end
+
+  private
+
+  def case_sensitive_labels?
+    false
+  end
 end


app/models/manageiq/providers/azure/cloud_manager.rb
```diff
@@ -45,6 +45,10 @@ def insights?
     end
   end
 
+  def case_insensitive_labels?
+    true
```
agrare, line 49: This should be `case_sensitive_labels? false`

chessbyte:
shouldn't the overridden methods here be `case_sensitive_labels?`

agrare (changes requested):

app/models/manageiq/providers/azure/cloud_manager.rb
lpichler: yes, it was leftover and I removed it now.

we have such method on azure's  persister

thanks

app/models/manageiq/providers/azure/inventory/persister.rb
```diff
@@ -23,4 +23,10 @@ def initialize(manager, target = nil, collector = nil)
 
     initialize_inventory_collections
   end
+
+  private
+
+  def case_sensitive_labels?
+    true
```
agrare, line 30: This should be false right?

lpichler:
> shouldn't the overridden methods here be `case_sensitive_labels?`

yes, it is fixed now 👍 thanks

agrare (approved):

app/models/manageiq/providers/azure/inventory/persister.rb
lpichler: yea good eye, it is fixed.

PR #424, opened 2020-11-05
Supports label mapping in cloud manager

lpichler:


app/models/manageiq/providers/azure/cloud_manager.rb
```diff
@@ -190,4 +192,16 @@ def snapshot_service(connection = nil)
   def self.display_name(number = 1)
     n_('Cloud Provider (Microsoft Azure)', 'Cloud Providers (Microsoft Azure)', number)
   end
+
+  LABEL_MAPPING_ENTITIES = {
+    "VmAzure" => "ManageIQ::Providers::Amazon::CloudManager::Vm"
```
chessbyte, line 197: typo: Amazon should be Azure

lpichler: fixed, thanks 👍

agrare (approved):

PR #425, opened 2020-12-14
private_constant rubocop fix

djberg96:
Fixes a minor rubocop complaint picked up by the linter. Kind of a weird place to do a splat and this is more readable anyway IMO.


agrare:
> Kind of a weird place to do a splat and this is more readable anyway IMO.

Yeah I agree `*[]` makes no sense

agrare (approved):

PR #430, opened 2021-01-25
adjust rel count in tests

d-m-u:
fix specs for the refactor of vms to use ancestry, not relationships

needs
https://github.com/ManageIQ/manageiq/pull/20788 

tested here: https://github.com/ManageIQ/manageiq-cross_repo-tests/pull/283


djberg96 (commented):
I know little of ancestry, but looks ok to me once the core PR is merged.

Issue #438, opened 2021-02-17
Azure all_valid_ems_in_zone does an Azure API call every sync_workers loop


agrare:
The Azure Provider only starts event_catchers if the `Microsoft.Insights` resource provider is registered.  This is done by calling [`insights?`](https://github.com/ManageIQ/manageiq-providers-azure/blob/master/app/models/manageiq/providers/azure/cloud_manager.rb#L42-L48) which is called by the [`supports?(:timeline)`](https://github.com/ManageIQ/manageiq-providers-azure/blob/master/app/models/manageiq/providers/azure/cloud_manager.rb#L28-L32) supports feature block.

The problem arises when the `Azure::CloudManager::EventCatcher` overrides [`all_valid_ems_in_zone`](https://github.com/ManageIQ/manageiq-providers-azure/blob/master/app/models/manageiq/providers/azure/cloud_manager/event_catcher.rb#L7-L13) to check `supports?(:timeline)`.

This looks innocuous but it is actually doing an Azure API call every time `Azure::CloudManager::EventCatcher.all_valid_ems_in_zone` is called.  This is done by the [`PerEmsWorkerMixin.sync_workers`](https://github.com/ManageIQ/manageiq/blob/master/app/models/mixins/per_ems_worker_mixin.rb#L32) method (by way of [`desired_queue_names`](https://github.com/ManageIQ/manageiq/blob/master/app/models/mixins/per_ems_worker_mixin.rb#L25-L27)).

Long story short we're doing an Azure API call from the MiqServer to check this feature every 15 seconds.

agrare:
https://github.com/ManageIQ/manageiq/issues/21061

Fryguy:
As a tactical fix, if we can't get all the capabilities work in time, I recommend using a cache_with_timeout of maybe 1.hour

jrafanie:
From chat with @agrare:

`insights?` is referenced as being a problem but, I'm finding that the time is pretty bad just connecting to the provider...

For example, here's what `insights?` looks like in the environment that opened this issue:

```
irb(main):003:0> ManageIQ::Providers::Azure::CloudManager.first.insights?
[...]

PR #439, opened 2021-02-17
Persist the Microsoft.Insights capability to the EMS#capabilities column

agrare:
Rather than perform an API call everytime we need to check if Microsoft.Insights is registered in this region we can save it to the EMS capabilities column during verify_credentials.

Depends on:
- [x] https://github.com/ManageIQ/manageiq-schema/pull/559

Fixes https://github.com/ManageIQ/manageiq-providers-azure/issues/438


Fryguy:
I don't understand how this code could have passed without the schema change merged.  I guess we don't have tests on this?

Fryguy (approved):

agrare:
:/ I guess not

Oh, it is stubbing the `ems.insights` call so it never hits `capabilities`.  I'll follow-up with a better test that sets the `capabilities` hash instead of mocking

PR #441, opened 2021-02-18
Replace ems.insights? stubs with setting capabilities

agrare:
Rather than stubbing the call to ems.insights? we can set the capabilities["insights"] attribute on the ems.

Follow-up to https://github.com/ManageIQ/manageiq-providers-azure/pull/439


djberg96 (approved):
LGTM

PR #450, opened 2021-03-31
Use pluggable provider secrets for vcr specs

agrare:


djberg96 (approved):
Yes, that looks better.

Issue #451, opened 2021-03-31
Refresh fails with error: "\xEF" from ASCII-8BIT to UTF-8, podified


evertmulder:
Hi,

when using the Azure provider from ManageIQ Kasparov-1 the refresh is always failing with this error:
`"\xEF" from ASCII-8BIT to UTF-8`. We are using the podified version running on Kubernetes.

I also tried the azure provider on a Jansa appliance. Here the refresh works without issues.
Then I also tried the azure provider on a Kasparov-1 appliance. Here the refresh works without issues.

I'm not sure what would be the difference between the podified and appliance, perhaps the logging configuration `/var/www/miq/vmdb/lib/vmdb/loggers/container_logger.rb:43`?

```
[----] I, [2021-03-30T21:14:17.610988 #7:2b25bdd0597c]  INFO -- : MIQ(ManageIQ::Providers::Azure::Inventory::Collector::CloudManager#collect_inventory) Retrieving Managed Images...
[----] I, [2021-03-30T21:14:17.663854 #7:2b25bdd0597c]  INFO -- : MIQ(ManageIQ::Providers::Azure::Inventory::Collector::CloudManager#collect_inventory) Retrieving Managed Images...Complete - Count [0]
[----] I, [2021-03-30T21:14:17.684629 #7:2b25bdd0597c]  INFO -- : MIQ(ManageIQ::Providers::Azure::Inventory::Collector::CloudManager#collect_inventory) Retrieving Private Images...
[----] E, [2021-03-30T21:14:17.980883 #7:2b25bdd0597c] ERROR -- : MIQ(ManageIQ::Providers::Azure::CloudManager::Refresher#refresh) EMS: [test2], id: [5] Refresh failed
[----] E, [2021-03-30T21:14:17.980991 #7:2b25bdd0597c] ERROR -- : [Encoding::UndefinedConversionError]: "\xEF" from ASCII-8BIT to UTF-8  Method:[block (2 levels) in <class:LogProxy>]
[----] E, [2021-03-30T21:14:17.981067 #7:2b25bdd0597c] ERROR -- : /opt/manageiq/manageiq-gemset/gems/json-2.5.1/lib/json/common.rb:312:in `generate'
/opt/manageiq/manageiq-gemset/gems/json-2.5.1/lib/json/common.rb:312:in `generate'
/var/www/miq/vmdb/lib/vmdb/loggers/container_logger.rb:43:in `call'
[...]

PR #453, opened 2021-04-03
Handle binary chars in private images exception message

agrare:
There is an exception message that contains an XML doc but starts with
some non-utf-8 characters which is causing the whole string encoding to
be ASCII 8-bit which Logger then fails to log.

```
xEF\xBB\xBF<?xml version=\"1.0\" encoding=\"utf-8\"?>
<Error><Code>TlsVersionNotPermitted</Code>...
```
Fixes https://github.com/ManageIQ/manageiq-providers-azure/issues/451

### DIFF
--- a/app/models/manageiq/providers/azure/inventory/collector/cloud_manager.rb
+++ b/app/models/manageiq/providers/azure/inventory/collector/cloud_manager.rb
@@ -35,7 +35,7 @@ class ManageIQ::Providers::Azure::Inventory::Collector::CloudManager < ManageIQ:
   def images
     collect_inventory(:private_images) { gather_data_for_this_region(@sas, 'list_all_private_images') }
   rescue ::Azure::Armrest::ApiException => err
-    _log.warn("Unable to collect Azure private images for: [#{@ems.name}] - [#{@ems.id}]: #{err.message}")
+    _log.warn("Unable to collect Azure private images for: [#{@ems.name}] - [#{@ems.id}]: #{err.message.force_encoding("utf-8")}")
     []
   end
 


agrare:
cc @djberg96 not sure if this has something to do with how azure-armrest is handling the exception or it is coming straight from azure this way

djberg96:
That's interesting, I don't think we do anything special with the exception handling.

I'm curious about the error itself, too. The azure-armrest gem uses TLSv1_2 by default.

djberg96 (approved):

Fryguy:
`xEF\xBB\xBF` is a [Byte-Order mark (BOM) for UTF-8](https://en.wikipedia.org/wiki/Byte_order_mark#UTF-8), so this makes sense.  I thought Ruby handled that automatically, or that there was a way to tell Ruby to honor the BOM.

djberg96:
One idea would be to modify azure-armrest's `Api::Exception` class (in exception.rb in the azure-armrest source) to automatically strip that in the constructor:
```ruby
# currently
def initialize(message = nil, cause_exception = nil)
  @message = message
  @cause = cause_exception
end

## CHANGE TO:
def initialize(message = nil, cause_exception = nil)
  @message = message.gsub!("\xEF\xBB\xBF".force_encoding("UTF-8"), '')
  @cause = cause_exception
end
```
Otherwise I would say it should be fixed in the base logger, which might be better since it would then work for any provider.

agrare:
There was some discussion here https://bugs.ruby-lang.org/issues/15210 about this but the patch that got applied looks specific to the `IO` https://github.com/ruby/ruby/commit/e717d6faa8463c70407e6aaf116c6b6181f30be6 which probably doesn't apply in this case since the exception string is probably not the first thing that we're getting back from the socket.

Fryguy:
@agrare Can you add a spec for this possibly?  That way when we see this code in the future we know why we added it :)

agrare:
@Fryguy yeah I'm still wondering if this is something we should handle generically in azure-armrest?  This one place might not be the only exception that has this.

Fryguy:
Oh yeah it definitely should be pushed down, if possible

Fryguy:
@agrare How do you want to proceed for now?  Do you want to push down, or should I just merge this one?

agrare:
@Fryguy I'm going to try handling this generically in the azure-armrest gem...will have a PR shortly

agrare:
Interestingly I just noticed that most of the responses returned are `application/json` but the one that fails is `application/xml`:
```
RestClient.post "https://management.azure.com/subscriptions, "", "Accept"=>"application/json", "Authorization"=>"Bearer [FILTERED] ", "Content-Length"=>"0", "Content-Type"=>"application/json", "User-Agent"=>"rest-client/2.1.0 (linux-gnu x86_64) ruby/2.6.3p62"
# => 200 OK | application/json 288 bytes, 0.11s
RestClient.get "https://csb10037ffea12a499b.blob.core.windows.net/?comp=list", "Accept"=>"*/*", "User-Agent"=>"rest-client/2.1.0 (linux-gnu x86_64) ruby/2.6.3p62", "auth_string"=>"true", "authorization"=>"SharedKey [FILTERED] ", "content-type"=>"", "x-ms-date"=>"Sat, 03 Apr 2021 17:28:04 GMT", "x-ms-version"=>"2016-05-31"
# => 400 BadRequest | application/xml 266 bytes, 0.09s
```

djberg96:
@agrare Yep, that's the way Azure setup unmanaged storage for some reason. Take a look at https://github.com/ManageIQ/azure-armrest/blob/master/lib/azure/armrest/model/storage_account.rb.

agrare:
@djberg96 PTAL https://github.com/ManageIQ/azure-armrest/issues/405 and https://github.com/ManageIQ/azure-armrest/pull/406

I'm going to close this since I think the issue is not properly handling xml responses generally in azure-armrest

PR #452, opened 2021-04-01
Replace stubbed authentication_status_ok? for event_catcher_spec

agrare:
Rather than stub ems.authentication_status_ok? and ems.supports?(:timeline) we can just create appropriate authentication records and set the capabilities

Required for: https://github.com/ManageIQ/manageiq/pull/21138

PR #456, opened 2021-04-27
[KASPAROV] Replace authentication_status_ok? with provider_authentication_status_ok?

agrare:
Kasparov specific version of https://github.com/ManageIQ/manageiq-providers-azure/pull/452 which cannot be backported due to the utilization of the new capabilities column

Depends:
- [x] https://github.com/ManageIQ/manageiq/pull/21198


agrare:
@Fryguy this is green

PR #459, opened 2021-06-14
Convert Vm#validate_pause to SupportsFeatureMixin

chessbyte:
Goes with https://github.com/ManageIQ/manageiq/pull/21278


djberg96:
Depends on how you want to define "pause". When you "stop" a VM in Azure, it still keeps some things allocated for it, like IP address and storage (which is presumably why you still get charged) which are retained if you hit "start" again. To actually "delete" a VM you have to deallocate it.

But no, Azure doesn't appear to have anything like pause/resume functionality like my local VM's with VMWare Fusion, for example. Just wanted to make sure it was clear.

chessbyte:
Thanks @djberg96 - this is just a refactoring.  It was not supported before in the `validate_pause` method and it is not supported after this PR is merged.

djberg96 (approved):

agrare (approved):

Issue #479, opened 2021-11-16
A failing Azure SSA scan doesn't delete the snapshot


agrare:
When an Azure SSA scan fails the `process_abort` doesn't cleanup the snapshot, causing future scans to fail with "Failed to create evm snapshot" "Another SSA request for this VM is in progress or a previous one failed to clean up properly."

PR #480, opened 2021-11-17
Fix Azure SSA not deleting snapshot

agrare:
The snapshot context wasn't being set in the scan_args which caused an exception to be raised in `perform_metadata_scan` [[here]](https://github.com/ManageIQ/manageiq-providers-azure/blob/master/app/models/manageiq/providers/azure/cloud_manager/vm_or_template_shared/scanning.rb#L43)

In addition the after_scan state was not implemented and thus the call_snapshot_delete state was never called.

TODO:
- [x] Clearly this needs some specs
- [x] The Azure override of `process_abort` isn't being called on failure leaving snapshots around

Fixes https://github.com/ManageIQ/manageiq/issues/21560
Fixes https://github.com/ManageIQ/manageiq-providers-azure/issues/479

### DIFF
--- a/app/models/manageiq/providers/azure/cloud_manager/scanning/job.rb
+++ b/app/models/manageiq/providers/azure/cloud_manager/scanning/job.rb
@@ -122,10 +122,10 @@ class ManageIQ::Providers::Azure::CloudManager::Scanning::Job < VmScan
   end
 
   # All other signals
-  alias_method :start_snapshot,  :call_snapshot_create
-  alias_method :snapshot_delete, :call_snapshot_delete
-  alias_method :abort_job,       :process_abort
-  alias_method :cancel,          :process_cancel
+  alias start_snapshot  call_snapshot_create
+  alias snapshot_delete call_snapshot_delete
+  alias abort_job       process_abort
+  alias cancel          process_cancel
 
   private
 
--- a/app/models/manageiq/providers/azure/cloud_manager/scanning/job.rb
+++ b/app/models/manageiq/providers/azure/cloud_manager/scanning/job.rb
@@ -93,8 +93,8 @@ class ManageIQ::Providers::Azure::CloudManager::Scanning::Job < VmScan
   def process_cancel(*args)
     begin
       delete_snapshot_and_reset_snapshot_mor("canceling")
-      super
-      rescue => err
+    rescue => err
+      _log.warn(err)
       _log.log_backtrace(err)
     end
 
@@ -104,8 +104,8 @@ class ManageIQ::Providers::Azure::CloudManager::Scanning::Job < VmScan
   def process_abort(*args)
     begin
       delete_snapshot_and_reset_snapshot_mor("aborting")
-      super
     rescue => err
+      _log.warn(err)
       _log.log_backtrace(err)
     end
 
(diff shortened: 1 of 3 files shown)


app/models/manageiq/providers/azure/cloud_manager/scanning/job.rb
```diff
@@ -167,4 +171,9 @@ def delete_snapshot_and_reset_snapshot_mor(log_verb)
     end
   end
 
+  def create_scan_args
+    super.tap do |scan_args|
+      scan_args["snapshot"] = {"name" => context[:snapshot_mor]}
```
agrare, line 176: TODO the snapshot "name" is only used by MiqAzureVm in the case of a "blob" snapshot.  The snapshot_mor returned by the azure create_snapshot method does appear to be the name but I need to confirm this

agrare: Okay confirmed the old code used the `context[:snapshot_mor]` as the `scan_args["snapshot"]["name"]` : https://github.com/ManageIQ/manageiq/pull/19607/files#diff-e8144513ccccfdc45e012a21875a9b40017058f6b00d0e4a4908326c245af0c9L202

app/models/manageiq/providers/azure/cloud_manager/scanning/job.rb
```diff
@@ -100,8 +104,8 @@ def process_cancel(*args)
   def process_abort(*args)
     begin
       delete_snapshot_and_reset_snapshot_mor("aborting")
-      super
```
agrare, line 103: This was calling `super` here and also down on like 112 at the end of the method, so it was failing saying can't transit state finished because it was already in state finished

app/models/manageiq/providers/azure/cloud_manager/scanning/job.rb
```diff
@@ -89,8 +93,8 @@ def call_snapshot_delete
   def process_cancel(*args)
     begin
       delete_snapshot_and_reset_snapshot_mor("canceling")
-      super
```
agrare, line 92: Deleted a duplicate `super` call, it was also being called down on line 101

app/models/manageiq/providers/azure/cloud_manager/scanning/job.rb
```diff
@@ -117,10 +121,11 @@ def snapshot_complete
     end
   end
 
-
   # All other signals
-  alias_method :start_snapshot,     :call_snapshot_create
-  alias_method :snapshot_delete,    :call_snapshot_delete
+  alias_method :start_snapshot,  :call_snapshot_create
+  alias_method :snapshot_delete, :call_snapshot_delete
+  alias_method :abort_job,       :process_abort
```
agrare, line 127: So unintuitive ruby behavior of the day, the `VmScan` class defines `alias_method :abort_job,       :process_abort`, this subclass overrides `process_abort` but `signal(:abort_job)` was calling `VmScan#process_abort` unless I also define the alias in the subclass :shrug:

app/models/manageiq/providers/azure/cloud_manager/scanning/job.rb
```diff
@@ -20,6 +20,10 @@ def before_scan
     signal(:start_snapshot)
   end
 
+  def after_scan
+    signal(:snapshot_delete)
+  end
```
agrare, line 25: This wasn't being overridden so the snapshot was never being deleted in the success case (which coincidentally never happened :laughing:)

agrare:
@Fryguy PTAL

Fryguy:
Backported to `morphy` in commit 021c076f7115a44a302b11800fbb45b3be66f082.

```text
commit 021c076f7115a44a302b11800fbb45b3be66f082
Date:   Mon Nov 22 09:52:08 2021 -0500

    Merge pull request #480 from agrare/fix_azure_ssa_not_deleting_snapshot
    
    Fix Azure SSA not deleting snapshot
    
    (cherry picked from commit 61af96a6c6142eed289dae66834f210e1926e36e)
```

PR #487, opened 2022-01-10
Add missing container_manager subclasses

agrare:
Required for: https://github.com/ManageIQ/manageiq/pull/21597


djberg96:
`ManageIQ::Providers::Kubernetes::ContainerManager::ContainerTemplate.include(ActsAsStiLeafClass)`

Kubernetes? Is that right?

agrare:
> Kubernetes? Is that right?

Yes, the Azure Kubernetes provider is a subclass of the Kubernetes provider.  `ActsAsStiLeafClass` allows us to inherit from other providers' models without the downsides of e.g. `ManageIQ::Providers::Kubernetes::ContainerManager::ContainerTemplate.all` including the `Azure` ones

djberg96 (approved):

PR #496, opened 2022-02-22
Fix self.verify_credentials region nil

agrare:
The region information passed to the class-level verify_credentials method from the DDF forms is in `args["provider_region"]` not `args["region"]`.

With a `nil` region you would see a `WARN -- azure: No region selected. Validating credentials against public environment.` warning in the logs and we would not verify credentials against the region selected.

### DIFF
--- a/app/models/manageiq/providers/azure/manager_mixin.rb
+++ b/app/models/manageiq/providers/azure/manager_mixin.rb
@@ -126,7 +126,7 @@ module ManageIQ::Providers::Azure::ManagerMixin
     #   }
     # }
     def verify_credentials(args)
-      region           = args["region"]
+      region           = args["provider_region"]
       subscription     = args["subscription"]
       azure_tenant_id  = args["uid_ems"]
       default_endpoint = args.dig("authentications", "default")


app/models/manageiq/providers/azure/manager_mixin.rb
```diff
@@ -126,7 +126,8 @@ def params_for_create
     #   }
     # }
     def verify_credentials(args)
-      region           = args["region"]
+      byebug
```
chessbyte, line 129:
```suggestion
```

Fryguy:
Backported to `najdorf` in commit fd64246b15ae448d8ed98c8b2d12568d08e53007.

```text
commit fd64246b15ae448d8ed98c8b2d12568d08e53007
Date:   Wed Feb 23 15:01:02 2022 -0500

    Merge pull request #496 from agrare/fix_verify_credentials_region
    
    Fix self.verify_credentials region nil
    
    (cherry picked from commit 21e64561a3230fb6a903f4c231e13859ef75dd33)
```

PR #497, opened 2022-02-22
Run an API call when verifying credentials

agrare:
Ensure that the VMM service is available in the region selected for the subscription provided.

Without this regions which we know are valid but not for certain subscriptions will return true for verify_credentials even when a subsequent refresh will fail.

For example I was able to successfully verify my credentials against "US Gov Virginia"

### DIFF
--- a/app/models/manageiq/providers/azure/manager_mixin.rb
+++ b/app/models/manageiq/providers/azure/manager_mixin.rb
@@ -11,16 +11,18 @@ module ManageIQ::Providers::Azure::ManagerMixin
   end
 
   def verify_credentials(_auth_type = nil, options = {})
-    conn = connect(options)
+    self.class.connection_rescue_block do
+      conn = connect(options)
 
-    # Check if the Microsoft.Insights Resource Provider is registered.  If not then
-    # neither events nor metrics are supported.
-    ms_insights_service = ::Azure::Armrest::ResourceProviderService.new(conn).get('Microsoft.Insights')
-    capabilities["insights"] = ms_insights_service.registration_state.casecmp('registered').zero?
+      # Check if the Microsoft.Insights Resource Provider is registered.  If not then
+      # neither events nor metrics are supported.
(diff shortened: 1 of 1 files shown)


app/models/manageiq/providers/azure/manager_mixin.rb
```diff
@@ -166,16 +173,14 @@ def raw_connect(client_id, client_key, azure_tenant_id, subscription, proxy_uri
 
       ::Azure::Armrest::Configuration.log = $azure_log
 
-      connection_rescue_block do
```
agrare, line 169: The rescue block seems to make more sense for `self.verify_credentials` where we need to format the response to be shown on the UI, raw_connect shouldn't mess with exceptions IMO (only used elsewhere by the instance-level connect)

app/models/manageiq/providers/azure/manager_mixin.rb
```diff
@@ -126,7 +128,7 @@ def params_for_create
     #   }
     # }
     def verify_credentials(args)
-      region           = args["region"]
+      region           = args["provider_region"]
```
Fryguy, line 131: was this a bug?

agrare: Yes, I have a separate PR to fix just this, the second commit in this branch builds on that fix

agrare: https://github.com/ManageIQ/manageiq-providers-azure/pull/496

Fryguy:
Backported to `morphy` in commit 57d1a4e7f88a1929c8ae8109f95b2871cb2636d6.

```text
commit 57d1a4e7f88a1929c8ae8109f95b2871cb2636d6
Date:   Wed Feb 23 15:19:47 2022 -0500

    Merge pull request #497 from agrare/run_simple_api_call_verify_credentials
    
    Run an API call when verifying credentials
    
    (cherry picked from commit 8e450435986b76d1457034b6c65a0cffdbd53264)
```

PR #499, opened 2022-03-21
Use supports?(:event) for all_valid_ems_in_zone

agrare:
The `:events` feature has been added to core in https://github.com/ManageIQ/manageiq/pull/21780 so we can be more specific with the event that we check here

### DIFF
--- a/app/models/manageiq/providers/azure/cloud_manager/event_catcher.rb
+++ b/app/models/manageiq/providers/azure/cloud_manager/event_catcher.rb
@@ -5,10 +5,6 @@ class ManageIQ::Providers::Azure::CloudManager::EventCatcher < ManageIQ::Provide
   # providers that do not support it.
   #
   def self.all_valid_ems_in_zone
-    super.select do |ems|
-      ems.supports?(:timeline).tap do |available|
-        _log.info(ems.unsupported_reason(:timeline) + " [#{ems.provider_region}]") unless available
-      end
-    end
+    super.select { |ems| ems.supports?(:events) }
   end
 end
--- a/app/models/manageiq/providers/azure/cloud_manager.rb
+++ b/app/models/manageiq/providers/azure/cloud_manager.rb
@@ -36,6 +36,12 @@ class ManageIQ::Providers::Azure::CloudManager < ManageIQ::Providers::CloudManag
   end
 
   supports :label_mapping
+
+  supports :events do
+    unless insights?
+      unsupported_reason_add(:events, _('Events are not supported for this region'))
+    end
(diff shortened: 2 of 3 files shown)


app/models/manageiq/providers/azure/cloud_manager/event_catcher.rb
```diff
@@ -5,10 +5,6 @@ class ManageIQ::Providers::Azure::CloudManager::EventCatcher < ManageIQ::Provide
   # providers that do not support it.
   #
   def self.all_valid_ems_in_zone
-    super.select do |ems|
-      ems.supports?(:timeline).tap do |available|
-        _log.info(ems.unsupported_reason(:timeline) + " [#{ems.provider_region}]") unless available
```
agrare, line 10: `all_valid_ems_in_zone` is called quite often by the `sync_workers` loop in MiqServer which means this gets logged very frequently if any EMSs don't support event capture.

I think we need a better way to bubble up why events can't be collected for a provider to the user than via the logs.

E.g. we could use the supports features to show this info on the UI on the provider dashboard.

app/models/manageiq/providers/azure/cloud_manager/event_catcher.rb
```diff
@@ -5,10 +5,6 @@ class ManageIQ::Providers::Azure::CloudManager::EventCatcher < ManageIQ::Provide
...
-      end
-    end
+    super.select { |ems| ems.supports?(:events) }
```
chessbyte, line 8: I expected the base class to have this code.

agrare: We could certainly do that if we go through and add `supports :events` to all providers which have an EventCatcher worker, however there aren't that many providers where collection of events is conditional at the instance level.  Most have event collection or they don't so just having an EventCatcher class is all that is required.

Unlike a lot of the core API calls where we're looking for all EMS classes which support a certain feature, we start out with the EMS worker class so if an EMS doesn't support events it won't have the class at all.

Fryguy:
Backported to `najdorf` in commit fea27431ec6912e20158ee7a9cbea2eae4f701e8.

```text
commit fea27431ec6912e20158ee7a9cbea2eae4f701e8
Date:   Tue Mar 22 15:50:58 2022 -0400

    Merge pull request #499 from agrare/use_supports_events
    
    Use supports?(:event) for all_valid_ems_in_zone
    
    (cherry picked from commit 8dc038da797cc4b9ce207ab2c2a8ded1f2ecc660)
```

PR #503, opened 2022-05-16
Improve Invalid Credentials error message

agrare:
There are four components that make up Azure credentials: tenant_id, subscription_id, client_id, and client_secret.

If you have the wrong subscription ID there is a different exception that is raised:
`#<ArgumentError: Subscription ID '2b01dec2-6684-4f25-abfc-a2453e0ae52e' not found>`

But if the Tenant ID, Client ID, or Client Secret are wrong we just get:
`#<Azure::Armrest::BadRequestException: [] Azure::Armrest::BadRequestException (cause: 400 Bad Request)>`

We were not indicating in the error message that the tenant ID could be what is wrong.

We are also not able to indicate _which_ of the three is wrong based on the error.

### DIFF
--- a/app/models/manageiq/providers/azure/manager_mixin.rb
+++ b/app/models/manageiq/providers/azure/manager_mixin.rb
@@ -191,7 +191,7 @@ module ManageIQ::Providers::Azure::ManagerMixin
     rescue ArgumentError => err
       raise MiqException::MiqInvalidCredentialsError, _("Incorrect credentials - %{error_message}") % {:error_message => err.message}
     rescue ::Azure::Armrest::UnauthorizedException, ::Azure::Armrest::BadRequestException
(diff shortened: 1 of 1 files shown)


djberg96 (approved):
Looks fine to me. I wonder if azure-armrest bubbles up a more detailed exception, or if Azure just isn't that granular with bad credential info.

Fryguy:
Backported to `najdorf` in commit ad65280ab2f6439bccf657a18d56a68ebf9092ad.

```text
commit ad65280ab2f6439bccf657a18d56a68ebf9092ad
Date:   Mon May 16 16:59:37 2022 -0400

    Merge pull request #503 from agrare/improve_invalid_credentials_error_message
    
    Improve Invalid Credentials error message
    
    (cherry picked from commit b270fa8072e55aafe3d494bcfa8478b37575e714)
```

PR #504, opened 2022-05-17
Create and delete cloud databases

nasark:
@miq-bot add_label enhancement
@miq-bot assign @agrare

### DIFF
--- a/app/models/manageiq/providers/azure/cloud_manager/cloud_database.rb
+++ b/app/models/manageiq/providers/azure/cloud_manager/cloud_database.rb
@@ -1,2 +1,103 @@
 class ManageIQ::Providers::Azure::CloudManager::CloudDatabase < ::CloudDatabase
+  supports :create
+  supports :delete
+
+  def self.params_for_create
+    {
+      :fields => [
+        {
+          :component => 'text-field',
+          :id        => 'name',
+          :name      => 'name',
+          :label     => _('Cloud Database Name'),
+        },
+        {
+          :component    => 'select',
+          :name         => 'database',
+          :id           => 'database',
+          :label        => _('Database Type'),
+          :includeEmpty => true,
+          :isRequired   => true,
+          :options      => ['MySQL', 'SQL', 'MariaDB', 'PostgreSQL'].map do |db|
+            {
+              :label => db,
+              :value => db
+            }
+          end,
+        },
+        {
+          :component => 'text-field',
+          :id        => 'server',
+          :name      => 'server',
+          :label     => _('Database Server Name'),
+        },
+        {
+          :component => 'text-field',
+          :id        => 'resource_group',
+          :name      => 'resource_group',
+          :label     => _('Resource Group'),
+        },
+      ],
+    }
+  end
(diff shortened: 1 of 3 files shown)


app/models/manageiq/providers/azure/cloud_manager/cloud_database.rb
```diff
@@ -1,2 +1,95 @@
...
+    ::Azure::Armrest::Sql::MysqlDatabaseService.new(ems.connect)
+  end
+
+  def self.get_postgresql_db_client(ems)
+    ::Azure::Armrest::Sql::PostgresqlDatabaseService.new(ems.connect)
+  end
+
+  def self.get_mariadb_db_client(ems)
+    ::Azure::Armrest::Sql::MariadbDatabaseService.new(ems.connect)
+  end
```
nasark, line 94: It would be nice if we could just use the methods found in https://github.com/ManageIQ/manageiq-providers-azure/blob/master/app/models/manageiq/providers/azure/refresh_helper_methods.rb#L229-275 to achieve the same purpose. However we would need to `extends ManageIQ::Providers::Azure::RefreshHelperMethods` and not sure if we want to be extending the module to the `CloudDatabase` class.

app/models/manageiq/providers/azure/cloud_manager/cloud_database.rb
```diff
@@ -1,2 +1,95 @@
...
+    db_client.create(options[:server], options[:name], options[:resource_group], {:location => ext_management_system.provider_region})
+  rescue => err
+    _log.error("cloud database=[#{options[:name]}], error: #{err}")
+    raise
+  end
+
+  def raw_delete_cloud_database
+    case db_engine
+    when /SQL Server/
+      self.class.get_sql_db_client(ext_management_system).delete_by_id(ems_ref)
```
nasark, line 67: These can be outsourced to their own individual methods i.e. `delete_sql_databases` but not sure if its necessary

app/models/manageiq/providers/azure/cloud_manager/cloud_database.rb
```diff
@@ -1,2 +1,95 @@
...
+    case options[:database]
+    when 'SQL'
+      db_client = get_sql_db_client(ext_management_system)
+    when 'MySQL'
+      db_client = get_mysql_db_client(ext_management_system)
+    when 'PostgreSQL'
+      db_client = get_postgresql_db_client(ext_management_system)
+    when 'MariaDB'
+      db_client = get_mariadb_db_client(ext_management_system)
+    end
```
agrare, line 56: Thinking we might want an
```ruby
else
  raise ArgumentError, etc...
```

If someone passes in an invalid option (e.g. post from the API without going through the UI)

app/models/manageiq/providers/azure/cloud_manager/cloud_database.rb
agrare: This looks simple enough

agrare (approved):

Issue #66, opened 2017-05-05
Add StorageManager?


djberg96:
Currently Azure supports two types of storage - Managed and Unmanaged, but we have no StorageManager to represent this in any way.

While we store disk information, we don't store overall storage information. This currently makes it difficult to distinguish VM's within the app that are managed versus those that are unmanaged.

Ideally, I should be able to tell by doing something like:

    vm.storage.store_type

Is this something we should add?

djberg96:
Closing for now.

PR #509, opened 2022-05-26
Delegate NetworkManager refresh to CloudManager

agrare:
Azure technically doesn't have the same bug as VPC since there is no storage manager and thus `manager or manager.network_manager` works, but still this is the direction we want to go in but not need to backport this


Fryguy (approved):

djberg96:
> Azure technically doesn't have the same bug as VPC since there is no storage manager

Alas, I never got around to it: https://github.com/ManageIQ/manageiq-providers-azure/issues/66

PR #512, opened 2022-07-05
Updating Create Cloud Database

MelsHyrule:
During testing for https://github.com/ManageIQ/manageiq-ui-classic/pull/8339 we found that all of these schema fields were needed to create a database. Additionally, `options` was expecting keys when we are passing in strings so that is also fixed here.

### DIFF
--- a/app/models/manageiq/providers/azure/cloud_manager/cloud_database.rb
+++ b/app/models/manageiq/providers/azure/cloud_manager/cloud_database.rb
@@ -2,14 +2,31 @@ class ManageIQ::Providers::Azure::CloudManager::CloudDatabase < ::CloudDatabase
   supports :create
   supports :delete
 
-  def self.params_for_create
+  def self.params_for_create(ems)
     {
       :fields => [
         {
-          :component => 'text-field',
-          :id        => 'name',
-          :name      => 'name',
-          :label     => _('Cloud Database Name'),
+          :component  => 'text-field',
+          :id         => 'name',
+          :name       => 'name',
+          :label      => _('Cloud Database Name'),
+          :isRequired => true,
+          :validate   => [{:type => 'required'}],
+        },
+        {
+          :component    => 'select',
+          :id           => 'resource_group',
+          :name         => 'resource_group',
+          :label        => _('Resource Group'),
+          :includeEmpty => true,
+          :isRequired   => true,
+          :validate     => [{:type => 'required'}],
+          :options      => ems.resource_groups.map do |rg|
+            {
+              :label => rg[:name],
+              :value => rg[:name],
+            }
+          end,
         },
         {
           :component    => 'select',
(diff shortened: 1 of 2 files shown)


MelsHyrule:
@miq-bot add-reviewer @jrafanie

app/models/manageiq/providers/azure/cloud_manager/cloud_database.rb
```diff
@@ -31,18 +33,25 @@ def self.params_for_create
...
+          :label      => _('Resource Group'),
+          :isRequired => true,
+          :validate   => [{:type => 'required'}],
         },
       ],
     }
   end
 
   def self.raw_create_cloud_database(ext_management_system, options)
+    options.symbolize_keys!
```
agrare, line 54: @MelsHyrule probably better to convert all of the `options[:key]` to `options["key"]` below rather than symbolizing keys here since we know the API will pass in string keys

djberg96 (approved):
Other than @agrare's suggestion, LGTM.

app/models/manageiq/providers/azure/cloud_manager/cloud_database.rb
```diff
@@ -31,18 +33,25 @@ def self.params_for_create
...
+          :isRequired => true,
+          :validate   => [{:type => 'required'}],
         },
         {
-          :component => 'text-field',
-          :id        => 'resource_group',
-          :name      => 'resource_group',
-          :label     => _('Resource Group'),
+          :component  => 'text-field',
+          ## TODO: Conver this into a drop down, dynamically populate it
```
jrafanie, line 41: typo:  convert

app/models/manageiq/providers/azure/cloud_manager/cloud_database.rb
jrafanie: Ok, so this means that line 55, 68 and the `db_client.create` call on 68 would need to use string keys and hopefully they work with them.

app/models/manageiq/providers/azure/cloud_manager/cloud_database.rb
```diff
@@ -18,32 +35,47 @@ def self.params_for_create
...
+        #       :value => svr[:name],
+        #     }
+        #   end,
+        #   :condition => {
+        #     :and => [
+        #       {when: 'resource_group', isNotEmpty: true,},
+        #       {when: 'database', isNotEmpty: true,},
+        #     ]
+        #   },
+        # },
```
agrare, line 72: Don't think we should to leave this code commented out here since the `ems.connect()` is not valid, if anything a TODO or NOTE above the database server name selection above that we should limit by the selected resource_group in the future

spec/models/manageiq/providers/azure/cloud_manager/cloud_database_spec.rb
```diff
@@ -72,10 +72,10 @@
 
     it 'creates a MySQL database' do
       expect(db_client).to receive(:create).with("test-server", cloud_database.name, "test-group", {:location => ems.provider_region})
-      cloud_database.class.raw_create_cloud_database(ems, {:name           => cloud_database.name,
-                                                           :server         => "test-server",
-                                                           :database       => "MySQL",
-                                                           :resource_group => "test-group"})
+      cloud_database.class.raw_create_cloud_database(ems, {"name"           => cloud_database.name,
```
agrare, line 75:
```suggestion
      cloud_database.class.raw_create_cloud_database(ems, {"name"           => cloud_database.name,
```

spec/models/manageiq/providers/azure/cloud_manager/cloud_database_spec.rb
```diff
@@ -85,7 +85,7 @@
     end
   end
 
-  describe 'postgresql database actions' do
+  describe 'postgresql database actions' do
```
agrare, line 88:
```suggestion
  describe 'postgresql database actions' do
```

spec/models/manageiq/providers/azure/cloud_manager/cloud_database_spec.rb
```diff
@@ -96,10 +96,10 @@
 
     it 'creates a PostgreSQL database' do
       expect(db_client).to receive(:create).with("test-server", cloud_database.name, "test-group", {:location => ems.provider_region})
-      cloud_database.class.raw_create_cloud_database(ems, {:name           => cloud_database.name,
-                                                           :server         => "test-server",
-                                                           :database       => "PostgreSQL",
-                                                           :resource_group => "test-group"})
+      cloud_database.class.raw_create_cloud_database(ems, {"name"           => cloud_database.name,
```
agrare, line 99:
```suggestion
      cloud_database.class.raw_create_cloud_database(ems, {"name"           => cloud_database.name,
```

agrare:
@MelsHyrule just trailing whitespace issues, probably a good idea to configure whatever editor you have to strip those on save

agrare (approved):

PR #513, opened 2022-07-07
Cloud database servers refresh support

nasark:
i.e.
```
 #<ManageIQ::Providers::Azure::CloudManager::CloudDatabaseServer:0x00007febc4205f78
  id: 9,
  name: "miq-sql-server",
  ems_ref:
   "/subscriptions/2b01dec2-6684-4f25-abfc-a2453e0ae52f/resourceGroups/test-group/providers/Microsoft.Sql/servers/miq-sql-server",
  ems_id: 15,
  server_type: "SQL",
  region: "East US",
  status: "Ready",
  version: "12.0",
  resource_group_id: 6>,
```

Dependent:
- [x] https://github.com/ManageIQ/manageiq/pull/21977
- [x] https://github.com/ManageIQ/manageiq-schema/pull/653

@miq-bot assign @agrare 
@miq-bot add_label enhancement

PR #514, opened 2022-07-15
Re-introducing Database Server Name drop-down

MelsHyrule:
With the merging of https://github.com/ManageIQ/manageiq-providers-azure/pull/513 we can now get the list of Database Server Names. Currently there isn't a way to make dependancies across DDF fields without doing something really hacky and specific to each of the provider, so for now it will be on the user to select the correct/corresponding server to their chosen `resource group` and `database type`.


app/models/manageiq/providers/azure/cloud_manager/cloud_database.rb
```diff
@@ -43,8 +43,22 @@ def self.params_for_create(ems)
...
+          :name         => 'server',
+          :id           => 'server',
+          :label        => _('Database Server Name'),
+          :includeEmpty => true,
+          :isRequired   => true,
+          :validate     => [{:type => 'required'}],
+          :options      => ems.cloud_database_servers.map do |srv|
+            {
+              :label => srv.name,
+              :value => srv.name,
```
agrare, line 58: @nasark just want to confirm that it is the server name which is meant to be passed to the Azure API (and not e.g. the ems_ref)

nasark: Yes, the server name is passed to the Azure API when creating a cloud database. For example:
```
> ManageIQ::Providers::Azure::CloudManager::CloudDatabase.raw_create_cloud_database(ems, {:database => "SQL", :name => "test", :resource_group => "test-group", :server => "miq-sql-server"})
```

nasark (approved):

agrare (approved):

Issue #400, opened 2020-06-22
Targeted refresh fails to save with `[TypeError]: can't cast Hash` failure


agrare:
```
[----] I, [2020-06-22T05:41:01.202632 #4738:2ae52a6d3960]  INFO -- : Q-task_id([r16_miq_provision_16]) MIQ(ManageIQ::Providers::Azure::CloudManager::Refresher#refresh) Refreshing all targets...
[----] I, [2020-06-22T05:41:01.202787 #4738:2ae52a6d3960]  INFO -- : Q-task_id([r16_miq_provision_16]) MIQ(ManageIQ::Providers::Azure::CloudManager::Refresher#refresh) EMS: [MSAzure], id: [255] Refreshing target
s for EMS...
[----] I, [2020-06-22T05:41:01.202937 #4738:2ae52a6d3960]  INFO -- : Q-task_id([r16_miq_provision_16]) MIQ(ManageIQ::Providers::Azure::CloudManager::Refresher#refresh) EMS: [MSAzure], id: [255]   InventoryRefres
h::TargetCollection [Collection of 1 targets] id [Collection of targets with id: [{:ems_ref=>{:subscription_id=>"0e0a4287-8719-4849-bb0b-5242e4507709", :vm_resource_group=>"infra-mgmt-auto-test", :type=>"microso
ft.compute/virtualmachines", :vm_name=>"test-ci-t1vp"}}]]
[----] I, [2020-06-22T05:41:13.493839 #4738:2ae52a6d3960]  INFO -- : Q-task_id([r16_miq_provision_16]) Exception in realtime_block :ems_refresh - Timings: {:collect_inventory_for_targets=>12.290756225585938, :em
s_refresh=>12.290783166885376}
[----] E, [2020-06-22T05:41:13.494053 #4738:2ae52a6d3960] ERROR -- : Q-task_id([r16_miq_provision_16]) MIQ(ManageIQ::Providers::Azure::CloudManager::Refresher#refresh) EMS: [MSAzure], id: [255] Refresh failed
[----] E, [2020-06-22T05:41:13.494187 #4738:2ae52a6d3960] ERROR -- : Q-task_id([r16_miq_provision_16]) [TypeError]: can't cast Hash  Method:[block (2 levels) in <class:LogProxy>]
[...]

PR #515, opened 2022-08-04
Fix `can't cast hash` error during provisioning

agrare:
When provisioning an instance a targeted refresh is queued using the value returned by the `start_clone` method as the `:ems_ref`.

Azure had been returning a hash rather than a string which was causing the Refresher to fail when querying for existing records using this hash as the `:ems_ref`.

The solution is to simply build the ems_ref the same way the Azure refresher does by joining the subscription, resource_group, type, and vm name.

Fixes https://github.com/ManageIQ/manageiq-providers-azure/issues/400
Fixes https://github.com/ManageIQ/manageiq/issues/21965

### DIFF
--- a/app/models/manageiq/providers/azure/cloud_manager/provision/cloning.rb
+++ b/app/models/manageiq/providers/azure/cloud_manager/provision/cloning.rb
@@ -1,17 +1,12 @@
 module ManageIQ::Providers::Azure::CloudManager::Provision::Cloning
-  def do_clone_task_check(clone_task_ref)
+  def do_clone_task_check(_clone_task_ref)
     source.with_provider_connection do |azure|
       vms      = ::Azure::Armrest::VirtualMachineService.new(azure)
-      instance = vms.get(clone_task_ref[:vm_name], clone_task_ref[:vm_resource_group])
+      instance = vms.get(dest_name, resource_group.name)
       status   = instance.properties.provisioning_state
       return true if status == "Succeeded"
       return false, status
     end
-  end
-
-  def find_destination_in_vmdb(vm_uid_hash)
-    ems_ref = vm_uid_hash.values.join("/")
-    ManageIQ::Providers::Azure::CloudManager::Vm.find_by("lower(ems_ref) = ?", ems_ref.downcase)
   end
 
   def gather_storage_account_properties
@@ -186,12 +181,7 @@ module ManageIQ::Providers::Azure::CloudManager::Provision::Cloning
       vms = ::Azure::Armrest::VirtualMachineService.new(azure)
       vm  = vms.create(dest_name, resource_group.name, clone_options)
 
-      {
-        :subscription_id   => azure.subscription_id,
-        :vm_resource_group => vm.resource_group,
-        :type              => vm.type.downcase,
-        :vm_name           => vm.name
-      }
+      File.join(azure.subscription_id, vm.resource_group.downcase, vm.type.downcase, vm.name)
     end
   end
 end
--- a/spec/models/manageiq/providers/azure/cloud_manager/provision_spec.rb
+++ b/spec/models/manageiq/providers/azure/cloud_manager/provision_spec.rb
@@ -21,30 +21,6 @@ describe ManageIQ::Providers::Azure::CloudManager::Provision do
       allow(subject).to receive(:create_nic).and_return(nic_id)
     end
 
-    context "#find_destination_in_vmdb" do
-      vm_uid_hash = {
-        :subscription_id => subscription_id,
-        :resource_group  => resource_group,
-        :type            => type,
-        :name            => name
-      }
-      it "VM in same sub-class" do
-        vm_uid_hash[:subscription_id] = subscription_id
-        vm
-        expect(subject.find_destination_in_vmdb(vm_uid_hash)).to eq(vm)
-      end
-
-      it "VM in same sub-class with invalid parameters" do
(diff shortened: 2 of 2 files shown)


app/models/manageiq/providers/azure/cloud_manager/provision/cloning.rb
```diff
@@ -186,12 +186,7 @@ def start_clone(clone_options)
       vms = ::Azure::Armrest::VirtualMachineService.new(azure)
       vm  = vms.create(dest_name, resource_group.name, clone_options)
 
-      {
-        :subscription_id   => azure.subscription_id,
-        :vm_resource_group => vm.resource_group,
-        :type              => vm.type.downcase,
-        :vm_name           => vm.name
-      }
+      File.join(azure.subscription_id, vm.resource_group.downcase, vm.type.downcase, vm.name)
```
agrare, line 189: NOTE this same logic lives in 3 different places now, need to refactor to reduce the duplication

app/models/manageiq/providers/azure/cloud_manager/provision/cloning.rb
```diff
@@ -1,19 +1,14 @@
...
       return true if status == "Succeeded"
       return false, status
     end
   end
 
-  def find_destination_in_vmdb(vm_uid_hash)
-    ems_ref = vm_uid_hash.values.join("/")
-    ManageIQ::Providers::Azure::CloudManager::Vm.find_by("lower(ems_ref) = ?", ems_ref.downcase)
-  end
-
```
agrare, line 16: Now that we return a normal ems_ref string this override is unnecessary (also it was missing the :ems_id :scream:)

Here is the core method: https://github.com/ManageIQ/manageiq/blob/4d1900133517c09d90df6a0712aa6ba776a0c9c9/app/models/manageiq/providers/cloud_manager/provision/cloning.rb#L2-L4

spec/models/manageiq/providers/azure/cloud_manager/provision_spec.rb
```diff
@@ -21,30 +21,6 @@
...
-        vm_uid_hash[:subscription_id] = "invalid_subscription_id"
-        expect(subject.find_destination_in_vmdb(vm_uid_hash)).to be_nil
-      end
-
-      it "VM in different sub-class" do
-        vm = FactoryBot.create(:vm_openstack, :ext_management_system => provider)
-        expect(subject.find_destination_in_vmdb(:ems_ref => vm.ems_ref)).to be_nil
-      end
-    end
-
```
agrare, line 47: NOTE now that we're using the method from the base class no need to test this here.

app/models/manageiq/providers/azure/cloud_manager/provision/cloning.rb
```diff
@@ -1,19 +1,14 @@
 module ManageIQ::Providers::Azure::CloudManager::Provision::Cloning
-  def do_clone_task_check(clone_task_ref)
+  def do_clone_task_check(_clone_task_ref)
     source.with_provider_connection do |azure|
       vms      = ::Azure::Armrest::VirtualMachineService.new(azure)
-      instance = vms.get(clone_task_ref[:vm_name], clone_task_ref[:vm_resource_group])
```
agrare, line 5: This was the only consumer of the values in the hash that was returned by `start_clone` and the values it needs are easily gotten from the Provision instance anyway.

agrare:
WIP pending a live test

app/models/manageiq/providers/azure/cloud_manager/provision/cloning.rb
Fryguy: Feels like it could be a method on the vm itself taking the subscription_id as a param.

agrare: Well `vm` here is the azure object not a `VmOrTemplate` instance since this is right after creating and before the record is created

djberg96: If all you need is a string, why not just `vm.id`?

agrare: This method is supposed to return the new vm's ems_ref so this has to match https://github.com/ManageIQ/manageiq-providers-azure/blob/master/app/models/manageiq/providers/azure/inventory/parser/cloud_manager.rb#L67-L70 exactly

agrare:
Confirmed fix with a live test:
`Refreshing target InventoryRefresh::TargetCollection [Collection of 8 targets] id [Collection of targets with id: [{:ems_ref=>"a6763ff5-a1f6-4f8c-885d-d2d1597f1921/agrare/microsoft.compute/virtualmachines/ag-prov-test-1"}`

agrare:
Going to tackle the refactoring in a follow-up to simplify backport, @Fryguy ready for review

Fryguy:
Backported to `najdorf` in commit 5ceac6deb1211a64d79ff799a5fef3bdb3ed86ac.

```text
commit 5ceac6deb1211a64d79ff799a5fef3bdb3ed86ac
Date:   Fri Aug 5 14:55:06 2022 -0400

    Merge pull request #515 from agrare/fix_azure_vm_provision_targeted_refresh
    
    Fix `can't cast hash` error during provisioning
    
    (cherry picked from commit efa2390a61270e2ce8423ba5863156a09ccf815d)
```

PR #516, opened 2022-08-10
Use lazy_find :key to replace a direct `find()`

agrare:
Rather than use a direct `.find()` which introduces ordering issues use the `:key` feature of lazy_find to get the `:device` after the lazy_find is evaluated

### DIFF
--- a/app/models/manageiq/providers/azure/inventory/parser/network_manager.rb
+++ b/app/models/manageiq/providers/azure/inventory/parser/network_manager.rb
@@ -290,14 +290,12 @@ class ManageIQ::Providers::Azure::Inventory::Parser::NetworkManager < ManageIQ::
     collector.floating_ips.each do |ip|
       uid = ip.id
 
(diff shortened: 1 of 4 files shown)


app/models/manageiq/providers/azure/inventory/parser/network_manager.rb
```diff
@@ -290,14 +290,12 @@ def floating_ips
...
       uid = ip.id
 
-      # TODO(lsmola) get rid of all the find method that are ineffective, a lazy_find multi should solve it
       network_port_id = floating_ip_network_port_id(ip)
-
-      network_port = persister.network_ports.find(network_port_id)
-      if network_port
-        vm = network_port.try(:[], :device)
-      elsif persister.load_balancers.find(network_port_id)
+      if load_balancer_network_port?(network_port_id)
```
agrare, line 294: We were using `.find()` to see if this was a VM or a LoadBalancer network port.  This is easily detected by checking the ID and we can avoid the ordering dependency.

PR #522, opened 2022-10-03
Remove useless gather_data_for_this_region method

agrare:
This method is a fairly useless wrapper which simply sends the method_name and didn't really share any code to filter by region anyway.

The `else` condition was added to handle stacks, but stack collection has been rewritten anyway and no longer uses this method

### DIFF
--- a/app/models/manageiq/providers/azure/inventory/collector.rb
+++ b/app/models/manageiq/providers/azure/inventory/collector.rb
@@ -104,15 +104,15 @@ class ManageIQ::Providers::Azure::Inventory::Collector < ManageIQ::Providers::In
   end
 
   def network_ports
-    @network_interfaces ||= collect_inventory(:network_ports) { gather_data_for_this_region(@nis) }
+    @network_interfaces ||= collect_inventory(:network_ports) { filter_my_region(@nis.list_all) }
   end
 
   def network_routers
-    @network_routers ||= collect_inventory(:network_routers) { gather_data_for_this_region(@rts) }
+    @network_routers ||= collect_inventory(:network_routers) { filter_my_region(@rts.list_all) }
   end
 
   def floating_ips
-    @floating_ips ||= collect_inventory(:floating_ips) { gather_data_for_this_region(@ips) }
+    @floating_ips ||= collect_inventory(:floating_ips) { filter_my_region(@ips.list_all) }
   end
 
   def instance_network_ports(instance)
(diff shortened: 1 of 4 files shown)


app/models/manageiq/providers/azure/inventory/collector.rb
```diff
@@ -262,14 +262,14 @@ def instances
   # that it doesn't affect the rest of inventory collection.
   #
   def images
-    collect_inventory(:private_images) { gather_data_for_this_region(@sas, 'list_all_private_images') }
+    collect_inventory(:private_images) { @sas.list_all_private_images(:location => @ems.provider_region) }
```
Fryguy, line 265: Does this need a filter_my_region call?

app/models/manageiq/providers/azure/inventory/collector/target_collection.rb
```diff
@@ -194,7 +194,7 @@ def images
 
     set = Set.new(refs)
 
-    collect_inventory_targeted(:private_images) { gather_data_for_this_region(@sas, 'list_all_private_images') }.select do |image|
+    collect_inventory_targeted(:private_images) { @sas.list_all_private_images(:location => @ems.provider_region) }.select do |image|
```
Fryguy, line 197: same here

app/models/manageiq/providers/azure/inventory/collector.rb
agrare: No because this call accepts a `:location` option directly we don't need it, see the old call https://github.com/ManageIQ/manageiq-providers-azure/pull/522/files#diff-8af42162f4f68708edd4edd030961533485f12f82576d34691847b9cef1f91dcL73-L74

Fryguy:
This particular rubocop catch is interesting:

**app/models/manageiq/providers/azure/inventory/collector.rb**
- [ ] :exclamation: - [Line 107](https://github.com/agrare/manageiq-providers-azure/blob/763173f39b4ba3beca901e83518bef39f6a5df96/app/models/manageiq/providers/azure/inventory/collector.rb#L107), Col 5 - [Naming/MemoizedInstanceVariableName](https://docs.rubocop.org/rubocop/1.28/cops_naming.html#namingmemoizedinstancevariablename) - Memoized variable `@network_interfaces` does not match method name `network_ports`. Use `@network_ports` instead.

Wonder if we should correct that in a followup.  It could be problematic in the future if we actually do create a network_interfaces method and accidentally the ivars clash.

agrare:
Yeah agreed, it looks like it should work since they both set `@network_interfaces` to `@nis.list_all` but very confusing.  The `get_network_interfaces` method is only used by the targeted_collector so we could move that over to calling `network_ports` directly and rename the ivar

Fryguy (approved):

PR #525, opened 2022-10-13
Add updated market_image_uris for newer versions

agrare:
Update the list of marketplace VM image URNs to represent newer versions of these operating systems.

Marketplace images aren't collected by default because there are over 32k of them, but the default list of image URN filters should be updated if someone wants to collect just these

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -50,9 +50,17 @@
     # Optionally collecting private images [Graph refresh only]
     :get_private_images: true
     :market_image_urns:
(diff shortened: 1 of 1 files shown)


Fryguy:
How many images come back with the filters in place?

agrare:
> How many images come back with the filters in place?

With filters in place?  The number of items in the filters list (it isn't * it is an exact URN) so as of now 11

Fryguy:
Ohhh I see :+1:

PR #526, opened 2022-10-13
Collect filtered marked images by default

agrare:
There are over 30k maket images so it isn't feasible to refresh all of them but we have a filtered list of the popular operating systems and recent versions so by default only the images in the filtered list are saved.

Follow-up to: https://github.com/ManageIQ/manageiq-providers-azure/pull/525#issuecomment-1277825018


Fryguy (approved):

agrare:
Just need to update the vcr counts

PR #543, opened 2023-10-31
Fix InventoryRefresh lazy_find deprecation warning

agrare:


agrare:
cc @kbrock fixing deprecation warnings feels like it is up your alley ;)

PR #549, opened 2024-02-23
drop unsupported_reason_add

kbrock:
part of:
- https://github.com/ManageIQ/manageiq/pull/22898
 
Deprecating `unsupported_reason_add`. Returning the string instead

Note, these are all the same:

```ruby
unsupported_reason_add(:feature, unsupported_reason(:feature2)) unless supports_feature2?
unsupported_reason_add(:feature, unsupported_reason(:feature2)) unless supports?(:feature2)
unsupported_reason(:feature2) unless supports?(:feature2)
unsupported_reason(:feature2)
```


app/models/manageiq/providers/azure/cloud_manager/vm/operations/power.rb
```diff
@@ -3,8 +3,8 @@ module ManageIQ::Providers::Azure::CloudManager::Vm::Operations::Power
 
   included do
     supports :reboot_guest do
-      unsupported_reason_add(:reboot_guest, unsupported_reason(:control)) unless supports?(:control)
-      unsupported_reason_add(:reboot_guest, _("The VM is not powered on")) unless current_state == "on"
+      unsupported_reason(:control) ||
+        (_("The VM is not powered on") unless current_state == "on")
```
agrare, line 7: The `|| (... unless condition)` works but looks overly complicated (I missed the added `()` at first)
What about using a ternary operator here? e.g.: `current_state == "on" ? unsupported_reason(:control) : _("The VM is not powered on")`
I assume `unsupported_reason(:control)` will return `nil` if it is supported right?

agrare (approved):

kbrock:
update:
- fixed `check_feature_support`
- fixed `unsupported_reason(:control)`

app/models/manageiq/providers/azure/cloud_manager/vm/operations/power.rb
kbrock: of note, this is technically different.
if current_state is off, and unsupported reason is false, then the error message coming out will be different.

But that is only a test failure issue, not a real issue.

app/models/manageiq/providers/azure/cloud_manager/vm/operations/power.rb
```diff
@@ -3,8 +3,11 @@ module ManageIQ::Providers::Azure::CloudManager::Vm::Operations::Power
 
   included do
     supports :reboot_guest do
-      unsupported_reason_add(:reboot_guest, unsupported_reason(:control)) unless supports?(:control)
-      unsupported_reason_add(:reboot_guest, _("The VM is not powered on")) unless current_state == "on"
+      if current_state == "on"
+        unsupported_reason(:control)
+      else
+        (_("The VM is not powered on") )
```
agrare, line 9: Good catch rubocop 
```suggestion
        _("The VM is not powered on")
```

kbrock:
update:
- removed unnecessary parens

agrare (approved):

PR #560, opened 2024-09-04
Clean up network interface and public ip on failure

liu-samuel:
When vm provision fails, there was a bug where the network interface and public ip address were still present of the ui console. This manually deletes them.

Azure has a mandatory 3 minute timeout after creation before i can delete the partial data so I'm sleeping for 180 seconds for now and I'm not sure how to get around it.

EDIT: Changed the sleep 180 to call requeue phase to add delete_instance back onto the queue

Related: 
https://github.com/ManageIQ/manageiq/pull/23175

### DIFF
--- a/app/models/manageiq/providers/azure/cloud_manager/provision/cloning.rb
+++ b/app/models/manageiq/providers/azure/cloud_manager/provision/cloning.rb
@@ -176,6 +176,16 @@ module ManageIQ::Providers::Azure::CloudManager::Provision::Cloning
     network_options
   end
 
+  def start_clone_task
+    super
+  rescue Azure::Armrest::BadRequestException => err
+    phase_context[:exception_class]     = err.class.name
+    phase_context[:exception_message]   = err.message
+    phase_context[:exception_backtrace] = err.backtrace
+    phase_context[:error_phase]         = phase
+    signal :clone_failure_cleanup
+  end
+
   def start_clone(clone_options)
     source.with_provider_connection do |azure|
       vms = ::Azure::Armrest::VirtualMachineService.new(azure)
--- a/app/models/manageiq/providers/azure/cloud_manager/provision/state_machine.rb
+++ b/app/models/manageiq/providers/azure/cloud_manager/provision/state_machine.rb
@@ -6,4 +6,24 @@ module ManageIQ::Providers::Azure::CloudManager::Provision::StateMachine
 
     signal :post_create_destination
   end
+
+  def clone_failure_cleanup
+    return if phase_context[:clone_options].nil?
+    
+    delete_instance(phase_context[:clone_options]) if phase_context[:clone_options]
+    signal :provision_error
+  rescue Azure::Armrest::BadRequestException
+    requeue_phase(3.minutes)
+  end
+
+  def delete_instance(instance_attrs)
+    source.with_provider_connection do |azure|
+      nis = ::Azure::Armrest::Network::NetworkInterfaceService.new(azure)
+      ips = ::Azure::Armrest::Network::IpAddressService.new(azure)
+      ni = nis.get(instance_attrs[:name], resource_group.name)
+      ip = ips.get("#{instance_attrs[:name]}-publicIp", resource_group.name)
+      nis.delete(instance_attrs[:name], resource_group.name) if ni
+      ips.delete("#{instance_attrs[:name]}-publicIp", resource_group.name) if ip
+    end
+  end
 end
--- a/spec/models/manageiq/providers/azure/cloud_manager/provision/cloning_spec.rb
+++ b/spec/models/manageiq/providers/azure/cloud_manager/provision/cloning_spec.rb
@@ -151,11 +151,13 @@ describe ManageIQ::Providers::Azure::CloudManager::Provision::Cloning do
       expect(ip_address_service).to have_received(:delete)
       expect(nic_service).to have_received(:delete)      
     end
+
     it 'phase_context is requeued properly after BadRequestException' do
       allow(nic_service).to receive(:delete).and_raise(Azure::Armrest::BadRequestException.new('errors', 'details', 'info'))
       provision.start_clone_task
       expect(provision).to have_received(:requeue_phase).with(3.minutes)
     end
+
     it 'phase_context errors are properly set after BadRequestException' do
       allow(nic_service).to receive(:delete).and_raise(Azure::Armrest::BadRequestException.new('errors', 'details', 'info'))
       provision.start_clone_task
--- a/spec/models/manageiq/providers/azure/cloud_manager/provision/cloning_spec.rb
+++ b/spec/models/manageiq/providers/azure/cloud_manager/provision/cloning_spec.rb
@@ -1,10 +1,18 @@
 require 'azure-armrest'
 
 describe ManageIQ::Providers::Azure::CloudManager::Provision::Cloning do
-  let(:provision) { ManageIQ::Providers::Azure::CloudManager::Provision.new }
+  let(:template) { FactoryBot.create(:azure_image, :ext_management_system => ems) }
+  let(:provision) { FactoryBot.create(:miq_provision_azure, :options => {:src_vm_id => template.id}, :miq_request => miq_request) }
+  let(:configuration) { double("azure armrest configuration") }
+  let(:vms) { double("virtual machine service") }
   let(:ems) { FactoryBot.create(:ems_azure_with_authentication) }
   let(:ip_address_service) { double('ip address service') }
+  let(:public_ip) { double('public ip') }
+  let(:network_interface) { double('network interface') }
+  let(:nic_service) { double("nic service") }
   let(:nic_id) { '/subscriptions/xyz/resourceGroups/foo/providers/Microsoft.Network/networkInterfaces/foo_nic' }
+  let(:miq_request) { FactoryBot.create(:miq_provision_request) }
+
 
   context "associated_nic" do
     before do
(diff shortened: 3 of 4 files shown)


app/models/manageiq/providers/azure/cloud_manager/provision/state_machine.rb
```diff
@@ -6,4 +6,22 @@ def customize_destination
...
+    delete_instance(phase_context[:clone_options])
+  end
+
+  def delete_instance(instance_attrs)
+    source.with_provider_connection do |azure|
+      nis = ::Azure::Armrest::Network::NetworkInterfaceService.new(azure)
+      ips = ::Azure::Armrest::Network::IpAddressService.new(azure)
+      ni = nis.get(instance_attrs[:name], resource_group.name)
+      ip = ips.get("#{instance_attrs[:name]}-publicIp", resource_group.name)
+      sleep(180)
```
Fryguy, line 22: Wonder if there's a way we can leverage the queue and say *run this 3 minutes from now"

app/models/manageiq/providers/azure/cloud_manager/provision/cloning.rb
```diff
@@ -183,5 +183,9 @@ def start_clone(clone_options)
 
       File.join(azure.subscription_id, vm.resource_group.downcase, vm.type.downcase, vm.name)
     end
+  rescue StandardError => e
+    return if phase_context[:clone_options].nil?
+    signal :clone_failure_cleanup
+    raise
```
agrare, line 189: `signal` is essentially a method call so this raise won't necessarily be reachable

app/models/manageiq/providers/azure/cloud_manager/provision/cloning.rb
```diff
@@ -183,5 +183,9 @@ def start_clone(clone_options)
 
       File.join(azure.subscription_id, vm.resource_group.downcase, vm.type.downcase, vm.name)
     end
+  rescue StandardError => e
+    return if phase_context[:clone_options].nil?
```
agrare, line 187: Could we move this check into `clone_failure_cleanup`

app/models/manageiq/providers/azure/cloud_manager/provision/cloning.rb
```diff
@@ -183,5 +183,9 @@ def start_clone(clone_options)
 
       File.join(azure.subscription_id, vm.resource_group.downcase, vm.type.downcase, vm.name)
     end
+  rescue StandardError => e
```
agrare, line 186: Can we rescue a more specific exception class?  Was there an AzureArmrest exception that was raised in this case?

app/models/manageiq/providers/azure/cloud_manager/provision/state_machine.rb
```diff
@@ -6,4 +6,21 @@ def customize_destination
 
     signal :post_create_destination
   end
+
+  def clone_failure_cleanup
+    delete_instance(phase_context[:clone_options])
```
agrare, line 11:
```suggestion
    delete_instance(phase_context[:clone_options]) if phase_context[:clone_options]
```

agrare: I think we should signal_abort after this is successful, so maybe something like
```suggestion
    begin
      delete_instance(phase_context[:clone_options])
      signal_abort
    rescue
      requeue_phase(3.minutes)
    end
```

app/models/manageiq/providers/azure/cloud_manager/provision/state_machine.rb
```diff
@@ -6,4 +6,21 @@ def customize_destination
...
+  def delete_instance(instance_attrs)
+    source.with_provider_connection do |azure|
+      nis = ::Azure::Armrest::Network::NetworkInterfaceService.new(azure)
+      ips = ::Azure::Armrest::Network::IpAddressService.new(azure)
+      ni = nis.get(instance_attrs[:name], resource_group.name)
+      ip = ips.get("#{instance_attrs[:name]}-publicIp", resource_group.name)
+      nis.delete(instance_attrs[:name], resource_group.name) if ni
+      ips.delete("#{instance_attrs[:name]}-publicIp", resource_group.name) if ip
+    end
+  rescue
```
agrare, line 23: Was there a specific exception that Azure raised when you weren't able to delete these ?

app/models/manageiq/providers/azure/cloud_manager/provision/cloning.rb
liu-samuel: Yeah, it was an Azure::Armrest::BadRequestException, so I can change it to 
`rescue Azure::Armrest::BadRequestException => e`

app/models/manageiq/providers/azure/cloud_manager/provision/state_machine.rb
liu-samuel: Yeah, this was also a Azure::Armrest::BadRequestException, so I can make it
`rescue Azure::Armrest::BadRequestException`

app/models/manageiq/providers/azure/cloud_manager/provision/state_machine.rb
agrare: > why do we need a requeue_phase here as well?

I think we should move the requeue_phase out of delete_instance and into here.  That way we keep all of the signal/requeue state directives in the same method

agrare:
Looks good @liu-samuel, I do think we need to add a spec test for this covering the various scenarios

app/models/manageiq/providers/azure/cloud_manager/provision/state_machine.rb
```diff
@@ -6,4 +6,23 @@ def customize_destination
 
     signal :post_create_destination
   end
+
+  def clone_failure_cleanup
+    return if phase_context[:clone_options].nil?
```
Fryguy, line 11: Minor but rubocop wants a blank line after guard clauses.

spec/models/manageiq/providers/azure/cloud_manager/provision/cloning_spec.rb
```diff
@@ -113,4 +116,33 @@
       expect(provision.create_nic(false)).to eql(@nic_object.id)
     end
   end
+
+  context "start_clone_task" do
+    before do
+      @resource_group = FactoryBot.create(:azure_resource_group)
+
+
```
Fryguy, line 124: blank lines? (accidental commit?)

agrare: I asked him to push up whatever he had so I could help him with his failing spec

spec/models/manageiq/providers/azure/cloud_manager/provision/cloning_spec.rb
```diff
@@ -1,10 +1,18 @@
 require 'azure-armrest'
 
 describe ManageIQ::Providers::Azure::CloudManager::Provision::Cloning do
-  let(:provision) { ManageIQ::Providers::Azure::CloudManager::Provision.new }
+  let(:template) { FactoryBot.create(:azure_image, ext_management_system: ems) }
```
agrare, line 4:
```suggestion
  let(:template) { FactoryBot.create(:azure_image, :ext_management_system => ems) }
```

spec/models/manageiq/providers/azure/cloud_manager/provision/cloning_spec.rb
```diff
@@ -1,10 +1,18 @@
...
+  let(:provision) { FactoryBot.create(:miq_provision_azure, :options => {:src_vm_id => template.id}, :miq_request => miq_request) }
+  let(:configuration) {double}
```
agrare, line 6:
```suggestion
  let(:configuration) {double('azure armrest configuration'}
```

Or something to indicate what the double is for

spec/models/manageiq/providers/azure/cloud_manager/provision/cloning_spec.rb
```diff
@@ -1,10 +1,18 @@
...
+  let(:vms) {double}
```
agrare, line 7: Same, something like 'virtual machine service'

spec/models/manageiq/providers/azure/cloud_manager/provision/cloning_spec.rb
```diff
@@ -113,4 +121,34 @@
       expect(provision.create_nic(false)).to eql(@nic_object.id)
     end
   end
+
+  context "start_clone_task" do
+    before do
+      @resource_group = FactoryBot.create(:azure_resource_group)
+      @dest_name = "test"
+      @clone_options = {clone_options: {name: "test-name", location: "germanycentralwest"}}
```
agrare, line 129: Same with these two if they aren't referenced elsewhere they can be simple local vars in your before block

spec/models/manageiq/providers/azure/cloud_manager/provision/cloning_spec.rb
```diff
@@ -113,4 +121,34 @@
...
+      allow(nic_service).to receive(:get).and_return(network_interface)
+      
+      allow(nic_service).to receive(:delete)
+      allow(ip_address_service).to receive(:delete)
+    end
+  
+    it 'deletes the public IP and network interface when a BadRequestException is raised' do
+      provision.start_clone_task
+      expect(ip_address_service).to have_received(:delete)
+      expect(nic_service).to have_received(:delete)
```
agrare, line 151: Nice

spec/models/manageiq/providers/azure/cloud_manager/provision/cloning_spec.rb
```diff
@@ -113,4 +121,34 @@
...
+      
+      allow(nic_service).to receive(:delete)
+      allow(ip_address_service).to receive(:delete)
+    end
+  
+    it 'deletes the public IP and network interface when a BadRequestException is raised' do
+      provision.start_clone_task
+      expect(ip_address_service).to have_received(:delete)
+      expect(nic_service).to have_received(:delete)      
+    end
```
agrare, line 152: @liu-samuel can you add a second `it` that checks the phase_context is set properly after BadRequestException is raised ?

agrare (approved):

Fryguy:
Backported to `radjabov` in commit 87c44f09bcaa077c229f241b9821a102d43422a9.

```text
commit 87c44f09bcaa077c229f241b9821a102d43422a9
Author: Adam Grare <adam@grare.com>
Date:   Tue Sep 10 17:54:35 2024 -0400

    Merge pull request #560 from liu-samuel/clean-up-failed-provision
    
    Clean up network interface and public ip on failure
    
    (cherry picked from commit 2c3a909ec3c7123422c641672c549d1f2d7aa0d2)
```

Issue #563, opened 2024-10-23
Refreshing Azure NetworkManager disconnects all managed_images


agrare:
```
INFO -- evm: MIQ(ManageIQ::Providers::Azure::NetworkManager::Refresher#refresh_targets_for_ems) EMS: [borazem Network Manager], id: [7] Refreshing target ManageIQ::Providers::Azure::NetworkManager [Azure Network Manager] id [7]...
INFO -- evm: MIQ(ManageIQ::Providers::Azure::Inventory::Parser::CloudManager#parse) Collecting data for EMS : [Azure Network Manager] id: [7]... 
INFO -- evm: MIQ(ManageIQ::Providers::Azure::Inventory::Collector::NetworkManager#collect_inventory) Retrieving Resource Groups...
INFO -- evm: MIQ(ManageIQ::Providers::Azure::Inventory::Collector::NetworkManager#collect_inventory) Retrieving Resource Groups...Complete - Count [2]
INFO -- evm: MIQ(ManageIQ::Providers::Azure::Inventory::Collector::NetworkManager#collect_inventory) Retrieving Series...
INFO -- evm: MIQ(ManageIQ::Providers::Azure::Inventory::Collector::NetworkManager#collect_inventory) Retrieving Series...Complete - Count [730]
INFO -- evm: MIQ(ManageIQ::Providers::Azure::Inventory::Collector::NetworkManager#collect_inventory) Retrieving Availability Zones...
INFO -- evm: MIQ(ManageIQ::Providers::Azure::Inventory::Collector::NetworkManager#collect_inventory) Retrieving Availability Zones...Complete - Count [1]
INFO -- evm: MIQ(ManageIQ::Providers::Azure::Inventory::Collector::NetworkManager#collect_inventory) Retrieving Deployments...
INFO -- evm: MIQ(ManageIQ::Providers::Azure::Inventory::Collector::NetworkManager#collect_inventory) Retrieving Deployments...Complete - Count [23]
INFO -- evm: MIQ(ManageIQ::Providers::Azure::Inventory::Collector::NetworkManager#stack_templates) Retrieving templates...
INFO -- evm: MIQ(ManageIQ::Providers::Azure::Inventory::Collector::NetworkManager#stack_templates) Retrieving templates...Complete - Count [0]
INFO -- evm: MIQ(ManageIQ::Providers::Azure::Inventory::Collector::NetworkManager#collect_inventory) Retrieving Instances...
[...]

PR #564, opened 2024-10-24
Fix Azure NetworkManager refresh disconnecting managed_images

agrare:
All refreshes should be done via the CloudManager, instance methods are delegated to the network_manager's parent_manager however the class-level refresh_ems method was not being delegated causing failures and disconnected inventory.

NOTE this was called via the scheduled `refresh_all_ems_timer` method, this currently goes through all permitted subclasses which includes the child managers like these Network and Storage managers under a parent cloud manager.  At best this is a duplicate full refresh and at worst in this case it exposes bugs like this.  I'd like to tackle running a full refresh on a timer only for the "primary" class (where that is going to differ provider to provider) in a separate pr.

Fixes https://github.com/ManageIQ/manageiq-providers-azure/issues/563

### DIFF
--- a/app/models/manageiq/providers/azure/network_manager.rb
+++ b/app/models/manageiq/providers/azure/network_manager.rb
@@ -25,6 +25,10 @@ class ManageIQ::Providers::Azure::NetworkManager < ManageIQ::Providers::NetworkM
            :to        => :parent_manager,
            :allow_nil => true
 
+  class << self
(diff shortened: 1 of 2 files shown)


agrare:
NOTE these changes are something which we should match in the other providers as well to prevent extra work from being done, but this is more important because here the collector settings cause images to be disconnected.

Fryguy (approved):

Fryguy:
Backported to radjabov via merge of master into radjabov

